PR #24, opened 2020-07-09
updated PHP and WordPress version

sekanderb:


sksaju:
readme file is now updated with your latest commit.

PR #32, opened 2020-10-23
Updated input fields with required and autocomplete

IvanRF:
Avoids form submit in modern browsers if the input fields are empty.

More info:
https://www.chromium.org/developers/design-documents/form-styles-that-chromium-understands
https://developer.mozilla.org/es/docs/Web/HTML/Atributos/autocomplete
https://developers.google.com/web/updates/2015/06/checkout-faster-with-autofill

### DIFF
--- a/templates/dashboard/instructor/registration.php
+++ b/templates/dashboard/instructor/registration.php
@@ -32,7 +32,7 @@
 					<?php _e('First Name', 'tutor'); ?>
                 </label>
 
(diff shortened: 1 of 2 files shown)


IvanRF:
Also, the input field for `email` should be of `type="email"`. I did not include that fix since some CSS changes are needed too. I leave that to you 😃

IvanRF:
Added another commit in order to fix dates localization in templates.

PR #33, opened 2020-10-29
Show dates in localized format

IvanRF:
Added another commit in order to fix dates localization in templates.


pmudry:
The idea of the commit is great. Maintainer should take a look at it... The dates look crap in french and german for instance.

PR #34, opened 2020-11-24
Fix infinity flush rewrite which is not required. Clean the queue right after flush rewrite once.

shohel:
**Issue**

- Install BuddyPress / BuddyBoss platform
- Install Tutor LMS
- Create some page from WordPress > pages > view
- Visit BuddyPress / BuddyBoss component page, like Timeline / Notification / Messages
- Return / Reload previously created the page and see 404 error

If you don't get any error, then repeat the same process and check the below steps.

- Visit BuddyPress / BuddyBoss component page, like Timeline / Notification / Messages
- Return / Reload previously created the page and see 404 error

**Solution**

This PR deleting the option `required_rewrite_flush`, _full line_ `delete_option( 'required_rewrite_flush' );` right after flush rewrite event, The flush should be rewrite only once when it require.


shohel:
@sksaju @sekanderb 

Check this PR, it will fix 404 page issue, solved conflict with **BuddyPress**, **BuddyBoss**

sksaju (approved):

sksaju (approved):

PR #64, opened 2021-11-02
Vimeo and Youtube video fetch duration api error fixed

Jobayer-Tuser:
First, we had HTTP and HTTPS set in the Vimeo and youtube API I have change the API please check


jayedul:
It can't be merged. You are not supposed to modify transpiled files.

PR #65, opened 2021-11-04
Fix Abandon Quiz? popup issue on Quiz page

creativeartbd:
If the quiz has pagination and if we click on the pagination link it's showing an Abandon Quiz popup message but it's shouldn't. So, I have added some JS logic here.


jayedul:
It can't be merged. Many things have been changed eventually.

PR #67, opened 2021-11-16
Add shortcode processing to external video allows fvplayer embed

bownie:
I wanted to use another plugin to play my video so added shortcode processing to the embed link. This works with fvplayer pro so that I can stream (e.g. HLS) signed URLs


jayedul:
Now Tutor LMS provides shortcode field natively along with other sources like embed code, youtube etc. Just need to enable from setting page.

PR #74, opened 2022-03-16
Option to Enforce registration (Create new account) on WooCommerce guest checkout

gerdneuman:
This implements the feature as described in https://wordpress.org/support/topic/enforce-registration-create-new-account-on-woocommerce-guest-checkout/

The problem like many others have pointed out in the [forum](https://wordpress.org/support/plugin/tutor/) before is that a guest needs to manually check the “Create a customer” account checkbox on checkout. This is often forgotten (by default it is also off).

It would be better if when a shopping cart includes a course, then the creation of a new account would be enforced on checkout (unless the user is already logged in).

### Technical details

The WooCommerce Subscriptions and WooCommerce Memberships plugins do it also this way using the same WooCommerce hooks. The PR uses the same approach:
 
* `woocommerce_checkout_registration_required` filter (which checks if cart contains a Tutor course, then yes) 
* and then also sets `$_POST['createaccount'] = 1;` using the `woocommerce_before_checkout_process` action in the network post request

### Some more word on why is this needed

For many, many shops like us, there is no point having a **guest** buying a course because how would a guest be able to access the course some days later if she/he has no acount and no login. Same reasoning exists for WooCommerce Subscriptions (for subscription products) and WooCommerce Memberships (also needs a user account). So guest checkout really makes no sense. Hence the option to enforce a registration/login during checking as done in the patch.  

Actually, in WooCommerce Subscriptions and WooCommerce Memberships this behavior is no option but enabled by default (because as said before there really is no point in a guest checkout (no account) for products like courses).

### DIFF
--- a/classes/Options_V2.php
+++ b/classes/Options_V2.php
@@ -683,6 +683,14 @@ class Options_V2
 								'desc'        => __('Allow customers to place orders without an account.', 'tutor'),
 							),
 							array(
+								'key'         => 'enforce_login_course_cart',
+								'type'        => 'toggle_switch',
+								'label'       => __( 'Enforce login or registration for courses on checkout', 'tutor' ),
+								'label_title' => __( '', 'tutor' ),
+								'default'     => 'off',
+								'desc'        => __( 'Enforces users to login or create a new account if cart contains a course. Disables guest checkout only if cart contains a course.', 'tutor' ),
+							),
+							array(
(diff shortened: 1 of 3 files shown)


gerdneuman:
@shewa12 This is closed why? If contributions are not welcome, that is fine, but better say so in the Readme then.

PR #103, opened 2022-04-08
 fix a bug that after finished paid but order status still pending

FeverChen:
tutor_utils()->get_option( 'auto_complete_woocommerce_virtual_orders' ) return true/false not 'on'/'off' and
 if not enabling auto_complete_woocommerce_virtual_orders, we still need to keep the woocommerce assigned order status, so return $payment_complete_status


slawekl:
I have the same problem and this fix resolved issue. Has been tested.

PR #180, opened 2022-08-19
Remove unused method and move utils method to model

harunollyo:
Remove unused methods from utils
```
    get_lesson_count
    get_lessons_by_topic
    get_lesson_type_icon
```

Move to LessonModel
```
    get_total_lesson
    get_lesson_count_by_course
    get_lesson_reading_info
    get_lesson_reading_info_full
```


sifat009:
I think `get_lesson_reading_info` has some dependencies.

PR #304, opened 2023-01-20
Fix: Initially all the topics of lesson sidebar remaining open (#1610)

shewa12:


templates/single/lesson/lesson_sidebar.php
```diff
@@ -94,7 +94,7 @@ function ( $content ) use ( $currentPost ) {
 				</div>
 			</div>
 
-			<div class="tutor-accordion-item-body" style="<?php echo $is_topic_active ? ' display: block;' : 'display: none;'; ?>">
+			<div class="tutor-accordion-item-body <?php echo $is_topic_active ? '' : 'tutor-display-none'; ?>">
```
harunollyo, line 97: Instead of making a new CSS class, we can re-use `tutor-d-none` class

PR #309, opened 2023-01-23
Fix - reveal mode not work for question below each other layout (1410)

harunollyo:


assets/react/front/course/_spotlight-quiz.js
```diff
@@ -10,7 +31,11 @@ window.jQuery(document).ready($=>{
 
         // Evaluate result
         var feedBackMode = $question_wrap.attr('data-quiz-feedback-mode');
```
shewa12, line 33: I think we should remove 
var feedBackMode = $question_wrap.attr('data-quiz-feedback-mode');
in the favour of is_reveal_mode()

harunollyo: done

PR #326, opened 2023-02-08
On the my profile page wrong registration time was appearing

wprashed:


templates/dashboard/my-profile.php
```diff
@@ -25,7 +25,7 @@
 $bio   = get_user_meta( $uid, '_tutor_profile_bio', true );
 
 $profile_data = array(
-	array( __( 'Registration Date', 'tutor' ), ( $rdate ? tutor_i18n_get_formated_date( $rdate ) : '' ) ),
```
harunollyo, line 28: We can use pre-defined tutor utils helper tutor_utils()->get_local_time_from_unix( $user-> user_registered )

PR #355, opened 2023-03-02
Fix: In the backend Q&A page, List Item design was not appearing

mi-prakash:
Student to Instructor chat issue
- As a student ask questions using Bullet Points
- As an instructor check in the backend that the Bullet Point is not appearing

After Fixing  
image

Task Url
https://ollyo.atlassian.net/browse/TUTOR-1491

### DIFF
--- a/assets/scss/modules/qna.scss
+++ b/assets/scss/modules/qna.scss
@@ -143,6 +143,18 @@
 					}
 				}
 			}
+
+			p {
+				margin: 0;
+			}
+
+			ul {
+				padding-left: 2rem;
+
+				li {
+					list-style-type: disc;
+				}
+			}
 		}
 	}
 
--- a/assets/scss/modules/qna.scss
+++ b/assets/scss/modules/qna.scss
@@ -261,7 +261,6 @@ table.qna-list-table {
 
 					li {
 						list-style-type: disc;
-						font-size: 13px;
     					line-height: 1.5;
 					}
 				}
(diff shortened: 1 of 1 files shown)


shewa12:
The issue is fixed but it would be great if you add below code snippet
```
ul {
	li {
		list-style-type: disc;
	}
}
```
inside this ```.tutor-qna-content line: 253 ```
So that on the Q & A page table could show formatted list like below:

image

mi-prakash:
> The issue is fixed but it would be great if you add below code snippet
> 
> ```
> ul {
> 	li {
> 		list-style-type: disc;
> 	}
> }
> ```
> 
> inside this `.tutor-qna-content line: 253 ` So that on the Q & A page table could show formatted list like below:
> 
> image

FIxed it. please check

assets/scss/modules/qna.scss
```diff
@@ -243,6 +255,16 @@ table.qna-list-table {
 				font-weight: bold;
 				font-size: 16px;
 				color: var(--tutor-body-color);
+
+				ul {
+					padding-left: 2rem;
+
+					li {
+						list-style-type: disc;
```
shewa12, line 263: @mi-prakash vai, any reason to specify ```font-size``` for list item? If we have text on a TD and a list item on another TD, it may look odd. If we don't specify then it will take font size according to the theme.

mi-prakash: ok removing the font size.
I used it because on the chat page font was used as 13 px on the list items

shewa12 (approved):

PR #381, opened 2023-03-23
Hooks to extend Tutor, reuse-able flash & alert message helpers 🎉 

shewa12:
The caching mechanism for flash message on top of `WP Transient`
Close-able alert message `tutor_close_able_alert_msg `

### DIFF
--- a/includes/tutor-general-functions.php
+++ b/includes/tutor-general-functions.php
@@ -1023,7 +1023,6 @@ if ( ! function_exists( 'tutor_meta_box_wrapper' ) ) {
 			);
 		}
 	}
-
 }
 
 if ( ! function_exists( 'tutor_close_able_alert_msg' ) ) {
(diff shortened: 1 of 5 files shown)


includes/tutor-general-functions.php
```diff
@@ -1023,5 +1023,29 @@ function( $classes ) use ( $custom_class ) {
...
+	 * Create a close-able alert message
+	 *
+	 * @since 2.1.9
+	 *
+	 * @param string $message alert message.
+	 * @param string $alert alert key like: success, warning, danger, etc.
+	 *
+	 * @return void
+	 */
+	function tutor_close_able_alert_msg( string $message, string $alert = 'success' ) {
```
harunollyo, line 1040: It'll nice if we use `closeable` as a single word.

includes/tutor-general-functions.php
```diff
@@ -1023,5 +1023,29 @@ function( $classes ) use ( $custom_class ) {
...
+		?>
+		<div class="tutor-alert tutor-<?php echo esc_attr( $alert ); ?> tutor-mb-12 tutor-alert tutor-success tutor-mb-12 tutor-d-flex tutor-align-center tutor-justify-between">
+			<span>
+				<?php echo esc_html( $message ); ?>
```
harunollyo, line 1044: We can allow basic HTML tags so that, we can show alert messages in various text formats.

harunollyo (approved):

PR #384, opened 2023-03-27
Security Enhancement 🎉 

shewa12:


classes/Reviews.php
```diff
@@ -45,6 +45,11 @@ public function delete_review() {
 		$review_id = Input::post( 'id', 0, Input::TYPE_INT );
 		$delete    = self::delete( $review_id );
 
+		// Check if user is privileged.
```
harunollyo, line 48: This check needs to be before the `self::delete` statement.

harunollyo (approved):

PR #402, opened 2023-04-05
Fixed Conflict with Other Plugin like Business Directory 🐞 

shewa12:


classes/Tutor.php
```diff
@@ -1075,7 +1075,7 @@ public static function create_withdraw_database() {
 	 * @return bool
 	 */
 	public function wp_doing_ajax( $bool ) {
-		if ( Input::has( 'tutor_ajax_action ' ) ) {
+		if ( isset( $_REQUEST['tutor_ajax_action'] ) ) {
```
harunollyo, line 1078: Add a comment here to protect this line from being changed.

PR #424, opened 2023-05-05
User helper method added

harunollyo:
Example
```php
use TUTOR\User;

User::is( User::ADMIN );

User::has_any_role( array( User::ADMIN, User::INSTRUCTOR ) );

User::is_admin( );
User::is_instructor( );
User::is_student( );
```

### DIFF
--- a/classes/User.php
+++ b/classes/User.php
@@ -123,14 +123,16 @@ class User {
 	}
 
 	/**
-	 * Check user is instructor.
+	 * Check current user is instructor.
 	 *
 	 * @since 2.2.0
 	 *
+	 * @param bool $is_approved instructor is approved or not.
+	 * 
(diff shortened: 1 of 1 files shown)


classes/User.php
```diff
@@ -49,6 +62,77 @@ public function __construct() {
...
+
+	/**
+	 * Check user is instructor.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @return boolean
+	 */
+	public static function is_instructor() {
+		return current_user_can( self::INSTRUCTOR );
```
shewa12, line 133: we can use `is_instructor` method from Utils class

shewa12 (approved):

shewa12 (approved):

PR #433, opened 2023-05-15
Date formate showing wrong after translation

Jobayer-Tuser:
The announcement publish date format is showing wrong if someone translating the tutor. I have modified the code now the translation will work in a proper format.

### DIFF
--- a/templates/single/course/enrolled/announcements.php
+++ b/templates/single/course/enrolled/announcements.php
@@ -36,7 +36,7 @@ $announcements = tutor_utils()->get_announcements( get_the_ID() );
 					</div>
 
 					<div>
-						<?php echo esc_html( human_time_diff( time(), strtotime( $announcement->post_date_gmt ) ) . __(' ago', 'tutor' ) ); ?>
+						<?php echo esc_html( sprintf( __( '%s ago', 'tutor' ), human_time_diff( strtotime(  $announcement->post_date_gmt ) ) ) ); ?>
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):
Agreed!

Being the acting translation editor of the German translation of the Tutor LMS plugin myself, I know all too well that it does not work for all languages to stick the translation of "ago" after the time diff.

Tutor LMS needs to use `"%s ago"` as translation string:

Only using `"%s ago"` allows the translation editors to use e.g. the translation `"vor %s"` for German.

Something like `"*timediff* her"` would be an approximation to keep the order, while that may be acceptable for times diffs which are larger than an hour, it's rather a clumsy translation if the time diff is below one hour.

Therefore, this fix is absolutely needed to allow translation editors to get the translations right.

Reference for `human_time_diff` showing that omitting the optional 2nd argument to `human_time_diff` is correct (the default for the 2nd argument is `time()`), so this pull request looks correct:
 
https://developer.wordpress.org/reference/functions/human_time_diff/

Example:
> https://plugins.trac.wordpress.org/browser/h5p/tags/1.14.1/admin/class-h5p-content-admin.php#L805:
> ```php
>    $human_time = sprintf(__('%s ago', $this->plugin_slug), $timediff);
> ```

@sifat009 @shewa12 @harunollyo Would you please merge this fix?

harunollyo (approved):

PR #442, opened 2023-05-17
Refactored WC product creation method 🎉 

shewa12:


models/QuizModel.php
```diff
@@ -754,4 +754,36 @@ public function get_first_or_last_attempt( int $quiz_id, int $user_id = 0, strin
...
+	 * @since 2.2.0
+	 *
+	 * @param int|array $course_id Course id or array of course ids.
+	 *
+	 * @return int
+	 */
+	public static function get_quiz_count_by_course( $course_id ) {
+		global $wpdb;
+
+		$and_clause = is_array( $course_id ) ? ' AND post_parent IN (' . implode( ',', $course_id ) . ')' : "AND post_parent = $course_id";
```
harunollyo, line 770: `prepare_in_clause` has a default sanitization check.
We can use `QueryHelper::prepare_in_clause( $course_id )` instead of `implode( ',', $course_id )`

classes/Utils.php
```diff
@@ -1495,14 +1495,20 @@ public function update_video( $post_id = 0, $video_data = array() ) {
 	 *
```
harunollyo, line 1495: Please another since tag after `@since 1.0.0` tag.

Like 
```
@since 1.0.0
@since 2.2.0 count param added to count attachment.
```

harunollyo (approved):

PR #479, opened 2023-06-15
HTML tags showing if user share the course on any social media

wprashed:


bernhardkaindl (changes requested):
@wprashed There are a few issues / TODOs:
- Please merge your commits into one commit
- Please fix the commit message to be descriptive of your change
- Please explain how you fix the fix: Not being an expert on that, know why the issue exists, and I do not understand how the use of the tutor_social_share_content filter fixes the issue.
- Please fix the return value issue shown in the "Checks" tab - why would this function suddenly have to have a return value?
- Please fix the white space isssues shown in the checks tab.

PR #480, opened 2023-06-15
Wrong enrollment count issue on the course list page.

wprashed:


templates/loop/meta.php
```diff
@@ -16,7 +16,7 @@
 $profile_url       = tutor_utils()->profile_url( $authordata->ID, true );
 $course_categories = get_tutor_course_categories( $course_id );
 $course_duration   = get_tutor_course_duration_context( $course_id, true );
-$course_students   = apply_filters( 'tutor_course_students', $course_id, tutor_utils()->count_enrolled_users_by_course( $course_id ) )
```
harunollyo, line 19: We can't remove `tutor_course_students` filter to solve the issue. The issue is not with the filter, issue is the return value of the filter.

wprashed: Bhai before we don't use any filter. It's common issue, every users are facing this issue if we are using this filter

bernhardkaindl: Well, at last Harun fixed the bug he introduced in #440 with where he messed up the order of arguments to apply_filters(), with #481. So the bug is fixed.

PR #493, opened 2023-07-06
Instructor reg phone number field optional and remove pattern from first and last name

harunollyo:


sifat009 (approved):
Looks good to me

PR #515, opened 2023-08-01
🎉 New: Private and Schedule filter tab in courses listing page.

harunollyo:


models/CourseModel.php
```diff
@@ -29,6 +29,8 @@ class CourseModel {
 	const STATUS_DRAFT      = 'draft';
 	const STATUS_AUTO_DRAFT = 'auto-draft';
 	const STATUS_PENDING    = 'pending';
+	const STATUS_PRIVATE    = 'private';
+	const STATUS_SCHEDULE   = 'future';
```
shewa12, line 33: I think the const name should have had`STATUS_FUTURE` since WP is calling it future.

views/pages/course-list.php
```diff
@@ -134,6 +133,10 @@
 	'private' => array( __( 'Private', 'tutor' ), 'select-default' ),
 );
 
+$future_list = array(
+	'publish' => array( __( 'Publish', 'tutor' ), 'select-success' ),
+	'future'  => array( __( 'Schedule', 'tutor' ), 'select-default' ),
```
shewa12, line 138: Label also should have `Future`

shewa12 (changes requested):
Please check recommendations

shewa12 (approved):

PR #517, opened 2023-08-02
Fix woocommerce deactivation handler

bernhardkaindl:
Hi! This fixes that the plugin shows a WooCommerce deactivation warning even when WooCommerce was not used when the WooCommerce deactivation handler runs:

The WooCommerce deactivation handler shall only act when `monetize_by === 'wc'`:

- Fix: When the WooCommerce plugin gets deactivated, and WooCommerce was not enabled in Tutor LMS:
  - Do not show a notice saying that paid courses would have been made free and to enable WooCommerce monetization to fix this.

- Fix: When the WooCommerce plugin gets deactivated, do not disable other possible monetization methods like Paid Memberships Pro (pmpro) or EDD.

### DIFF
--- a/classes/WooCommerce.php
+++ b/classes/WooCommerce.php
@@ -89,7 +89,7 @@ class WooCommerce extends Tutor_Base {
 		 * @since 1.7.8
 		 */
 		$woocommerce_path = dirname( dirname( __DIR__ ) ) . DIRECTORY_SEPARATOR . 'woocommerce' . DIRECTORY_SEPARATOR . 'woocommerce.php';
-		register_deactivation_hook( $woocommerce_path, array( $this, 'disable_tutor_monetization' ) );
+		register_deactivation_hook( $woocommerce_path, array( $this, 'woocommerce_deactivation_handler' ) );
 		/**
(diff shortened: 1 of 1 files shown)


bernhardkaindl:
@shewa12 Hi! Please have a look!

Even if WooCommerce Monetization is not used, when WooCommerce gets deactivated, Tutor LMS currently adds a permanent banner at the top of the admin dashboard which only gets cleared when WooCommerce gets re-activated.

The only trick to get around this right now is to deactivate tutor, then deactivate/uninstall WooCommerce, and the re-activate tutor again, and to get the idea for this trick, you have to study the source code deeply.

bernhardkaindl:
@shewa12 Why did you close this PR?

The problem that this change fixes is:

Even if WooCommerce Monetization is not used, when WooCommerce gets deactivated, Tutor LMS currently adds a permanent banner at the top of the admin dashboard which only gets cleared when WooCommerce gets re-activated.

The only trick to get around this right now is to deactivate tutor, then deactivate/uninstall WooCommerce, and the re-activate tutor again, and to get the idea for this trick, you have to study the source code deeply.

PR #419, opened 2023-04-19
adding support for PMPro

a11smiles:
While Tutor supports PMPro as a monetization engine, the current `get_orders_by_user_id` implementation returns an empty `orders` array.

I've updated (and tested) the function to return orders captured in PMPro.

Changes:
1. Added condition for PMPro and WooCommerce (both return `shop_order` for the post type).
2. Made the second inner join optional (for PMPro orders, the meta key `_is_tutor_order_for_course` isn't saved).

PR #523, opened 2023-08-06
Fix and improve support for PMPro (Paid Memberships Pro)

bernhardkaindl:
@shewa12 
While Tutor supports PMPro as a monetization engine, the current implementation of get_orders_by_user_id() returns an empty orders array.

Fix this function to return orders captured in PMPro, implemented and tested by Joshua Davis (@a11smiles)

- Add condition for PMPro and WooCommerce (both return shop_order for the post type).
- Add support for querying PMPro subscriptions
- Make the second inner join optional (for PMPro orders, the meta key _is_tutor_order_for_course isn't saved).

Was originally opened by Joshua Davis <me@jdav.is> @a11smiles as #419,
- commits and log messages squashed,
- rebased to current dev
- and spaces at end of lines fixed by me.

Originally-From: Joshua Davis <me@jdav.is>

### DIFF
--- a/classes/Utils.php
+++ b/classes/Utils.php
@@ -2658,6 +2658,49 @@ class Utils {
 	}
 
 	/**
+	 * @param $order_id
+	 *
+	 * @return array
+	 *
+	 */
+	public function get_subscriptions_by_order_id( $order_id ) {
+		global $wpdb;
+
+		// Return all children of a subscription
+		$args = array(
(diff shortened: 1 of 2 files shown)


bernhardkaindl:
@shewa12

It would be more friendly if you'd describe why you close PRs.

If you do not accept community contributions, please say so in the README.

PR #547, opened 2023-09-04
Blank lesson comment restriction

harunollyo:


classes/Lesson.php
```diff
@@ -563,8 +564,14 @@ public function tutor_lesson_completed_after( $content_id ) {
 	 */
 	public function reply_lesson_comment() {
 		tutor_utils()->checking_nonce();
+		$comment = Input::post( 'comment', '', Input::TYPE_KSES_POST );
+		if ( 0 === strlen( $comment ) ) {
+			wp_send_json_error();
```
shewa12, line 569: We  can send a message with `wp_send_json_error` to let the user know the exact reason for failure

assets/react/front/course/_lesson.js
```diff
@@ -9,10 +9,18 @@ window.jQuery(document).ready($=>{
...
     $(document).on('click', '.tutor-single-course-lesson-comments button[type="submit"]', function(e){
         e.preventDefault();
+
         const {__} = wp.i18n;
         let btn = $(this);
         let form = btn.closest('form');
         let data = form.serialize();
+        let comment = form.find('textarea[name="comment"]').val();
+        
+        if (comment.length === 0) {
```
shewa12, line 19: I think we should `trim` before checking to prevent empty spaces or tabs like below:
`comment.trim().length === 0`

shewa12 (changes requested):

shewa12 (approved):

PR #549, opened 2023-09-07
Course slug is now update-able 🎉 

shewa12:


bernhardkaindl (approved):
@shewa12 It looks like this should have been merged into 2.3.0 before release to update the course slug in the permalinks of the courses!

PS: The same update of the lesson permalinks seems to be missing when changing the lession_permalink_base!!!

bernhardkaindl (approved):
@shewa12: See #564, where I rebased this PR to the current dev (resolved the conflicts)

PR #564, opened 2023-09-22
classes/Options_V2: Update the permalink when changing the course slug

bernhardkaindl:
Also, fold a variable that was only used once.

This is @shewa12's #549, rebased (with conflicts resolved) to the current dev branch.

### DIFF
--- a/classes/Options_V2.php
+++ b/classes/Options_V2.php
@@ -636,6 +636,15 @@ class Options_V2 {
 						'block_type' => 'uniform',
 						'fields'     => array(
 							array(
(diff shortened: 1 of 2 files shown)


shewa12:
This feature is already released

bernhardkaindl:
> This feature is already released

But it is not renaming the course base slug of existing courses. These changes which were not releases could fix it, I guess?

bernhardkaindl:
@shewa12 I guess handling changes to the course and lesson base slugs may have been fixed meanwhile, but it would be great if you could communicate about it.

PR #565, opened 2023-09-27
🎉 New: Spotlight mode now features distinct icons for complete, pass, fail, and pending statuses.

harunollyo:
Related PR https://github.com/themeum/tutor-pro/pull/318

### DIFF
--- a/models/QuizModel.php
+++ b/models/QuizModel.php
@@ -20,6 +20,10 @@ use Tutor\Helpers\QueryHelper;
  */
 class QuizModel {
 
+	const ATTEMPT_STARTED = 'attempt_started';
+	const ATTEMPT_ENDED   = 'attempt_ended';
+	const REVIEW_REQUIRED = 'review_required';
+
 	/**
 	 * Get quiz table name
 	 *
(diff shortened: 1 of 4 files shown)


models/QuizModel.php
```diff
@@ -887,4 +891,50 @@ public static function get_quiz_count_by_course( $course_id ) {
...
+		$all_pending = true;
+		$result      = 'pending';
+
+		$user_id      = tutor_utils()->get_user_id( $user_id );
+		$attempt_list = $wpdb->get_results( $wpdb->prepare( "SELECT * FROM {$wpdb->prefix}tutor_quiz_attempts WHERE user_id=%d AND quiz_id=%d", $user_id, $quiz_id ) );
+
+		/**
+		 * Check all attempts are pending or not
+		 */
+		foreach ( $attempt_list as $attempt ) {
```
shewa12, line 917: Instead of looping to find out pending and calculate results, we can do it inside SQL. Here is an example:

`SELECT 
    attempt_id,
    user_id,
    quiz_id,
    total_marks,
    earned_marks,
    CAST((earned_marks / total_marks) * 100 AS UNSIGNED) AS percentage,
    (
        CASE 
            WHEN earned_marks >= (0.80 * total_marks) 
            THEN 'pass' 
            ELSE 'fail' 
        END
    ) AS result
FROM
    wp_tutor_quiz_attempts
WHERE
    attempt_status != 'review_required';
`

harunollyo: With this query, we can't figure out whether quiz result (Single/Multiple) attempts are pending state or not.

shewa12 (approved):

PR #566, opened 2023-09-27
Auto course complete restriction on strict mode if student not passed quiz or assignment.

harunollyo:
Included
   - **Update:** on auto course complete redirect to course details with review popup.

### DIFF
--- a/models/CourseModel.php
+++ b/models/CourseModel.php
@@ -133,7 +133,7 @@ class CourseModel {
 	 *
 	 * @since 1.0.0
 	 *
-	 * @param $instructor_id
+	 * @param int $instructor_id instructor ID.
 	 *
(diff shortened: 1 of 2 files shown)


models/CourseModel.php
```diff
@@ -580,4 +589,67 @@ public static function get_paid_courses( string $meta_key, array $args = array()
...
+					if ( 'pass' !== $result ) {
+						return false;
+					}
+				}
+			}
+
+			return true;
+		}
+
+		return false;
```
shewa12, line 652: Isn't this return statement redundant here?

shewa12 (approved):

models/CourseModel.php
harunollyo: No

PR #573, opened 2023-10-03
profile page social link mobile device issue -TUTOR-1978

ibrahim-kardi:
Social links aren't working on the Instructor profile page ( mobile view )
card link : https://ollyo.atlassian.net/browse/TUTOR-1978

### DIFF
--- a/templates/public-profile.php
+++ b/templates/public-profile.php
@@ -157,8 +157,21 @@ do_action( 'tutor_profile/' . $user_type . '/before/wrap' );
 				</div>
 
 				<div class="tutor-social-container content-for-mobile">
-					<?php echo wp_kses( $social_media, $allowed_html_for_social ); ?>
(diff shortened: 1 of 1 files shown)


templates/public-profile.php
```diff
@@ -157,9 +157,22 @@
 				</div>
 
 				<div class="tutor-social-container content-for-mobile">
-					<?php echo wp_kses( $social_media, $allowed_html_for_social ); ?>
```
harunollyo, line 160: Instead of removeing this line, just add `href` to `$allowed_html_for_social`

harunollyo (changes requested):

templates/public-profile.php
ibrahim-kardi: updated

harunollyo (approved):

PR #604, opened 2023-11-14
added usermeta for active users

ibrahim-kardi:


classes/User.php
```diff
@@ -62,6 +62,7 @@ public function __construct( $register_hooks = true ) {
 
 		add_action( 'admin_notices', array( $this, 'show_registration_disabled' ) );
 		add_action( 'admin_init', array( $this, 'hide_notices' ) );
+		add_action( 'tutor_after_login_success', array( $this, 'tutor_update_user_last_login' ) );
```
harunollyo, line 65: Not nessasary to use `tutor` prefix when use class method as callback.

classes/User.php
```diff
@@ -376,4 +377,15 @@ public function show_registration_disabled() {
...
 		<?php
 	}
+	/**
+	 * Set the user last active timestamp to now.
+	 *
+	 * @since 2.5.0
+	 * @param int $user_id User ID to check active user.
+	 */
+	public function tutor_update_user_last_login( $user_id ) {
+		update_user_meta( $user_id, 'tutor_last_login', (string) strtotime( gmdate( 'Y-m-d', time() ) ) );
```
harunollyo, line 387: save only timestamp value instead of formated date only

harunollyo (changes requested):

classes/User.php
```diff
@@ -376,4 +377,16 @@ public function show_registration_disabled() {
...
+	/**
+	 * Set the user last active timestamp to now.
+	 *
+	 * @since 2.5.0
+	 *
+	 * @param string $user_login active user name.
+	 * @param int    $user User ID to check active user.
+	 */
+	public function update_user_last_login( $user_login, $user ) {
+		update_user_meta( $user->ID, 'tutor_last_login', time() );
```
harunollyo, line 389: Use `current_time(  'timestamp',  true  )` instead of `time()`

harunollyo (changes requested):

harunollyo (approved):

PR #662, opened 2024-01-30
Tab component added with orientation

sazedul-haque:
Screenshot 2024-01-30 at 5 56 04 PM


assets/react/v3/entries/course-builder/pages/CourseBasic.tsx
```diff
@@ -145,4 +186,18 @@ const styles = {
     border-left: 1px solid ${colorTokens.stroke.default};
     min-height: calc(100vh - (${headerHeight}px + ${footerHeight}px));
   `,
+  courseSettings: css`
+    display: grid;
+    grid-template-columns: 200px 1fr;
+    margin-top: ${spacing[48]};
+    margin-bottom: ${spacing[48]};
```
ahamed, line 193:
```suggestion
    margin-block: ${spacing[48};
```

assets/react/v3/shared/config/styles.ts
```diff
@@ -418,6 +418,7 @@ export const spacing = {
   8: '8px',
   10: '10px',
   12: '12px',
+  14: '14px',
```
ahamed, line 421: This should not be added to our design system. If need to use some spacing out of the design system then use it as hard-coded or use nearby one.

ahamed (approved):

assets/react/v3/shared/config/styles.ts
sazedul-haque: Removed!

PR #685, opened 2024-02-13
fix(tutor_permission_denied_template): call die() instead of return to correctly end execution of code

### LINKED ISSUE #686: Unauthenticated users can see Course Builder
Hi all,

just formally creating an issue for my pull request #685  here as well.

When you are a regular user, without any special elevated rights, e.g. if you are a student and try to open up the course builder page, you will see something like the below:

https://demo.themeum.com/tutor/dashboard/create-course/?course_ID=7773

image

It loads the "permission denied template" twice, and then even proceeds to load the course builder template.

The user cannot actually save any changes, but they shouldn't be able to see any of that in the first place.

a fix is available via #685

pano9000:
The code was previously using to "return" at the end of the function to stop execution of code.

The way the `tutor_permission_denied_template` is used though, returning will not help -> you need to use die() to halt further execution.

e.g. you can see that issue here:
https://demo.themeum.com/tutor/dashboard/create-course/?course_ID=7773

The template is from `tutor-pro` / `frontend-course-builder.php`:

in there code does two checks, that are supposed to stop printing the course-builder template - however they use the `tutor_permission_denied_template` function, which just returns, so execution just goes on, and everything is printed anyways.
(e.g. it also therefore prints the "permission denied template" twice, because it is called twice.)

fixes #686


pano9000:
pinging @harunollyo @shewa12 to kindly get a closer look at this

PR #696, opened 2024-02-22
[FE] Improve the FormInput field and accept unit for displaying on the right

b-l-i-n-d:


sazedul-haque:
We have a separate input field for this. Please check `FormInputWithContent.tsx` component.

PR #701, opened 2024-02-28
[FE] Create the time limit field

b-l-i-n-d:
As we already have `FormInput` and `FormSelectInput`, this component can be created using those two. I have modified `FormInput` and `FormSelectInput` to receive `removeBorder` props to create the desired `TimeLimit` filed using those two components.


assets/react/v3/shared/components/fields/FormSelectInput.tsx
```diff
@@ -77,6 +80,10 @@ const FormSelectInput = <T,>({
     isDropdown: true,
   });
 
+  const attribute = {
```
ahamed, line 83:
```suggestion
  const additonalAttribute = {
```

ahamed (approved):

PR #729, opened 2024-03-13
Add the quiz modal layout with basic settings

ahamed:


assets/react/v3/shared/components/modals/ModalWrapper.tsx
```diff
@@ -175,7 +175,7 @@ const styles = {
     }
   `,
   content: css`
-    height: calc(100% - ${modal.HEADER_HEIGHT + modal.MARGIN_TOP}px);
+    height: 100%;
```
b-l-i-n-d, line 178: This will cause overflow not to act as expected. We have tried `height: 100%` before. This does not trigger overflow when it is supposed to. Also sometimes the content gets outside of the screen.

b-l-i-n-d (changes requested):

PR #733, opened 2024-03-14
Left side question list with mock content.

ahamed:


assets/react/v3/entries/course-builder/components/modals/QuizModal.tsx
```diff
@@ -234,6 +284,88 @@ const styles = {
     display: grid;
     grid-template-columns: 352px 1fr 352px;
     height: 100%;
+    background-color: ${colorTokens.surface.courseBuilder};
```
b-l-i-n-d, line 287: I think this should be added to the `ModalWrapper` content as all modals have the same `background-color`. Also, this will fix the current issue of background not applying to the overflowed section.

b-l-i-n-d (approved):

PR #743, opened 2024-03-20
[FE] Create an alert atom

b-l-i-n-d:
Issue: [Jira](https://ollyo.atlassian.net/browse/TUTOR-2181)


assets/react/v3/shared/atoms/Alert.tsx
```diff
@@ -0,0 +1,53 @@
...
+interface AlertProps {
+	children: React.ReactNode;
+	type?: AlertType;
+	icon?: IconCollection;
+}
+
+const Alert = ({ children, type = 'warning', icon }: AlertProps) => {
+	return (
+		<div css={styles.wrapper({ type })}>
+			<SVGIcon style={styles.icon({ type })} name={icon as IconCollection} height={24} width={24} />
```
ahamed, line 20: Why do I need to cast the type explicitly here?

assets/react/v3/shared/config/styles.ts
```diff
@@ -215,6 +215,13 @@ export const colorTokens = {
       onHold: '#ac0640',
       cancelled: '#6f7073',
     },
+    alert: {
+      warning: '#D47E00',
+      success: '#D47E00', // will be updated later
+      danger: '#D47E00', // will be updated later
+      info: '#D47E00', // will be updated later
+      primary: '#D47E00', // will be updated later
+    }
```
ahamed, line 224: This pattern does not align with our design system. If it is out of our design system then move these colors to the Alert component and define there.

assets/react/v3/shared/config/styles.ts
```diff
@@ -244,6 +251,13 @@ export const colorTokens = {
       cancelled: '#eceef2',
       refunded: '#e5f5f5',
     },
+    alert: {
+      warning: '#FBFAE9',
+      success: '#D47E00', // will be updated later
+      danger: '#D47E00', // will be updated later
+      info: '#D47E00', // will be updated later
+      primary: '#D47E00', // will be updated later
```
ahamed, line 259: Same here.

assets/react/v3/shared/config/styles.ts
```diff
@@ -262,6 +276,13 @@ export const colorTokens = {
     success: '#22a848',
     drop: '#4761b8',
     processing: '#00a388',
+    alert: {
+      warning: '#FAB000',
+      success: '#D47E00', // will be updated later
+      danger: '#D47E00', // will be updated later
+      info: '#D47E00', // will be updated later
+      primary: '#D47E00', // will be updated later
+    }
```
ahamed, line 285: Same here.

ahamed (changes requested):

assets/react/v3/shared/atoms/Alert.tsx
b-l-i-n-d: As icon is optional it can be `undefined` so we need to cast the type explicitly. I forgot to do the conditional check. I will fix it on the next commit.

PR #784, opened 2024-06-26
chore: Refactor code to calculate total children recursively

b-l-i-n-d:
Added `calculateTotalChildren` to calculate all nested children

### DIFF
--- a/assets/react/v3/shared/components/fields/FormCategoriesInput.tsx
+++ b/assets/react/v3/shared/components/fields/FormCategoriesInput.tsx
@@ -166,13 +166,12 @@ interface BranchProps {
   isLastChild: boolean;
 }
 
-// all children nested under this category
-const calculateTotalChildren = (option: CategoryWithChildren): number => {
-  return option.children.reduce((acc, child) => acc + calculateTotalChildren(child), option.children.length);
+const getTotalNestedChildrenCount = (option: CategoryWithChildren): number => {
+  return option.children.reduce((total, child) => total + getTotalNestedChildrenCount(child), option.children.length);
 };
 
 export const Branch = ({ option, value, onChange, isLastChild }: BranchProps) => {
-  const totalChildren = calculateTotalChildren(option);
+  const totalChildren = getTotalNestedChildrenCount(option);
   const hasChildren = totalChildren > 0;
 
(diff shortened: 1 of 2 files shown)


assets/react/v3/shared/components/fields/FormCategoriesInput.tsx
```diff
@@ -166,8 +166,13 @@ interface BranchProps {
   isLastChild: boolean;
 }
 
+// all children nested under this category
```
ahamed, line 169:
```suggestion
// Calculate all children nested under this category
```
Maybe we don't need to add comment instead make the function name more descriptive

assets/react/v3/shared/components/fields/FormCategoriesInput.tsx
```diff
@@ -166,8 +166,13 @@ interface BranchProps {
...
+const calculateTotalChildren = (option: CategoryWithChildren): number => {
+  return option.children.reduce((acc, child) => acc + calculateTotalChildren(child), option.children.length);
```
ahamed, line 171:
```suggestion
  return option.children.reduce((total, child) => total + calculateTotalChildren(child), option.children.length);
```

assets/react/v3/shared/utils/util.ts
```diff
@@ -256,20 +256,20 @@ export const parseNumberOnly = (value: string) => {
...
-		if (!inThrottle) {
-			func.apply(this, args);
-			inThrottle = true;
-
-			setTimeout(() => {
-				inThrottle = false;
-			}, limit);
-		}
-	};
+  let inThrottle = false;
```
ahamed, line 259: I don't find any usability of this variable.

ahamed (changes requested):

assets/react/v3/shared/components/fields/FormCategoriesInput.tsx
b-l-i-n-d: change the name to `getTotalNestedChildrenCount`

assets/react/v3/shared/utils/util.ts
b-l-i-n-d: `inThrottle` is used inside the return function

PR #801, opened 2024-07-02
Ecommerce Database Schema Design

shewa12:
Total 8 new tables has been introduced

### DIFF
--- a/classes/Tutor.php
+++ b/classes/Tutor.php
@@ -555,16 +555,23 @@ final class Tutor {
 		add_action( 'init', array( $this, 'init_action' ) );
 
 		/**
-		 * Redirect to the wizard page
+		 * Check activated plugin
 		 *
 		 * @since 1.5.7
 		 */
 
 		add_action( 'activated_plugin', array( $this, 'activated_tutor' ), 10, 2 );
+
+		/**
+		 * Redirect to setup page
+		 *
+		 * @since 2.8.0
+		 */
+		add_action( 'admin_init', array( $this, 'redirect_to_setup_page' ) );
 	}
 
 	/**
-	 * Redirect to the wizard page
+	 * Check activated plugin
 	 *
 	 * @since 1.5.7
 	 *
@@ -575,7 +582,19 @@ final class Tutor {
 	 */
 	public function activated_tutor( $plugin, $network_wide = null ) {
 		if ( tutor()->basename === $plugin ) {
-			if ( ( ! get_option( 'tutor_wizard' ) ) && version_compare( TUTOR_VERSION, '1.5.6', '>' ) ) {
+			$this->redirect_to_setup_page();
+		}
+	}
+	/**
+	 * Redirect to setup page
+	 *
+	 * @since 2.8.0
+	 *
+	 * @return void
+	 */
+	public function redirect_to_setup_page() {
+		if ( ( ! get_option( 'tutor_wizard' ) ) && version_compare( TUTOR_VERSION, '1.5.6', '>' ) ) {
+			if ( ! wp_doing_ajax() ) {
 				update_option( 'tutor_wizard', 'active' );
 				wp_safe_redirect( admin_url( 'admin.php?page=tutor-setup' ) );
 				exit;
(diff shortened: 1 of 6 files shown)


classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			discount_amount DECIMAL(13, 2),
+			discount_reason TEXT,
+			tax DECIMAL(13, 2) COMMENT 'Tax percentage',
+			tax_amount DECIMAL(13, 2),
+			fees DECIMAL(13, 2),
+			earnings DECIMAL(13, 2),
+			payment_method VARCHAR(255),
+			payment_payloads LONGTEXT,
+			note TEXT,
+			created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
harunollyo, line 878: - We should not use `CURRENT_TIMESTAMP` to these field `created_at` and `update_at`. Because `CURRENT_TIMESTAMP` use mysql server time where we need user site time.
- `created_at_gmt` and `updated_at_gmt` needed to store 0 offset GMT time.
- We should avoid use of `ENUM`

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
+			coupon_code VARCHAR(50) NOT NULL,
+			coupon_title VARCHAR(255) NOT NULL,
+			coupon_description TEXT,
+			discount_type ENUM('percentage', 'flat') NOT NULL,
+			discount_amount DECIMAL(13, 2) NOT NULL,
+			total_usage_limit INT(10) UNSIGNED,
+			per_user_usage_limit INT(10) UNSIGNED,
+			coupon_type VARCHAR(100) NOT NULL, -- product specific or category or all
+			utc_start_date DATETIME NOT NULL,
```
harunollyo, line 920: We should use `start_date_gmt` and `expire_date_gmt` instead of `utc_` prefix

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			utc_expire_date DATETIME DEFAULT NULL,
+			created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
harunollyo, line 922: - We should avoid use of `CURRENT_TIMESTAMP`

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			KEY user_id (user_id),
+			CONSTRAINT fk_tutor_coupon_usage_coupon_code FOREIGN KEY (coupon_code) REFERENCES {$wpdb->prefix}tutor_coupons(coupon_code) ON DELETE CASCADE,
+			CONSTRAINT fk_tutor_coupon_usage_user_id FOREIGN KEY (user_id) REFERENCES {$wpdb->prefix}users(ID) ON DELETE CASCADE
+		) $charset_collate;";
+
+		$cart_table = "CREATE TABLE {$wpdb->prefix}tutor_carts (
+			id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
+			user_id BIGINT(20) UNSIGNED DEFAULT NULL,
+			coupon_code VARCHAR(50) DEFAULT NULL,
+			created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
harunollyo, line 953: CURRENT_TIMESTAMP

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
 			PRIMARY KEY (withdraw_id)
 		) $charset_collate;";
 
+		$orders_table = "CREATE TABLE {$wpdb->prefix}tutor_orders (
+			id BIGINT(20) UNSIGNED AUTO_INCREMENT,
+			payment_id BIGINT(20) UNSIGNED COMMENT 'Unique payment id from payment gateway',
```
ahamed, line 861: Better use `transaction_id`

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			user_id BIGINT(20) UNSIGNED NOT NULL,
+			status ENUM('pending', 'completed', 'cancelled', 'refunded', 'partially-refunded') NOT NULL,
```
ahamed, line 863: This is hard to alter `ENUM` if you need to change the table structure. Better use varchar and use comment.

ahamed: We need to separate the payment status and the order status. One order could be `refunded` and could be `cancelled` at the same time.

ahamed: We may need to store the statuses like `pending, paid, failed, refunded, partially-refunded`

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			sub_total_price DECIMAL(13, 2) NOT NULL, -- price calculation based on course sale price
```
ahamed, line 864: `subtotal` is a single word.
```suggestion
			subtotal_price DECIMAL(13, 2) NOT NULL, -- price calculation based on course sale price
```

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			total_price DECIMAL(13, 2) NOT NULL,
+			order_price DECIMAL(13, 2) NOT NULL,
+			coupon_code VARCHAR(255),
+			discount_type ENUM('percentage', 'flat'),
+			discount_amount DECIMAL(13, 2),
+			discount_reason TEXT,
+			tax DECIMAL(13, 2) COMMENT 'Tax percentage',
```
ahamed, line 871: `tax_rate` is more discernible.

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			tax_amount DECIMAL(13, 2),
```
ahamed, line 872: Maybe `taxable_amount`

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			KEY order_id (order_id),
+			KEY course_id (course_id),
+			CONSTRAINT fk_tutor_order_item_order_id FOREIGN KEY (order_id) REFERENCES {$wpdb->prefix}tutor_orders(id) ON DELETE CASCADE
+		) $charset_collate;";
+
+		$coupons_table = "CREATE TABLE {$wpdb->prefix}tutor_coupons (
+			id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
+			coupon_code VARCHAR(50) NOT NULL,
+			coupon_title VARCHAR(255) NOT NULL,
+			coupon_description TEXT,
```
ahamed, line 914: I believe this is redundant to use `coupon_` prefix of the column names inside the `coupons` table.

classes/Tutor.php
```diff
@@ -837,13 +856,134 @@ public static function create_database() {
...
+			discount_type ENUM('percentage', 'flat') NOT NULL,
+			discount_amount DECIMAL(13, 2) NOT NULL,
+			total_usage_limit INT(10) UNSIGNED,
+			per_user_usage_limit INT(10) UNSIGNED,
+			coupon_type VARCHAR(100) NOT NULL, -- product specific or category or all
```
ahamed, line 919: Same here.

ahamed (changes requested):

classes/Tutor.php
shewa12: I believe, despite the table name, column names should be descriptive & it's the convention. For ex: on the wp_posts table there columns lie: post_title, post_content etc.

classes/Tutor.php
shewa12: updated

classes/Tutor.php
shewa12: updated

classes/Tutor.php
shewa12: Actually, tax_amount will be thrown from the payment gateway, it means tax has already been paid. That's why we kept the name `tax_amount`

PR #873, opened 2024-07-15
WPEditor component added

sazedul-haque:


views/modal/wp_editor_link.php
```diff
@@ -0,0 +1,62 @@
+<div id="wp-link-backdrop" style="display: none"></div>
```
harunollyo, line 1: Add this file comment

```
<?php
/**
 * WP Editor link modal
 *
 * @package Tutor\Views
 * @author Themeum <support@themeum.com>
 * @link https://themeum.com
 * @since 3.0.0
 */
```

harunollyo (changes requested):

views/modal/wp_editor_link.php
sazedul-haque: Added

PR #897, opened 2024-07-18
Add manual payment modal added with ajax integration

sazedul-haque:


classes/Options_V2.php
```diff
@@ -1760,4 +1760,25 @@ public function view_template( $template_slug, $section = array() ) {
...
 	}
+
+	/**
+	 * Save tutor manual payments.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return void send wp_json response
+	 */
+	public function add_manual_payment_method() {
```
shewa12, line 1771: Move this method to the `ecommerce/settings`

views/options/settings.php
```diff
@@ -134,5 +134,6 @@
 	<?php
 	//phpcs:ignore -- contain safe data
 	echo $this->view_template( 'common/modal-confirm.php', array() );
+	require_once tutor()->path . 'views/modal/ecommerce/add-manual-payment.php';
```
shewa12, line 137: Include modal if `tutor_utils()->is_monetize_by_tutor()` true

shewa12 (changes requested):
Please resolve the suggested changes

PR #908, opened 2024-07-23
Cart and Checkout pages functionalities added

sazedul-haque:


sifat009 (approved):

ecommerce/CartController.php
```diff
@@ -94,7 +94,7 @@ public function __construct( $register_hooks = true ) {
 			 *
 			 * @since 3.0.0
 			 */
-			add_action( 'wp_ajax_tutor_cart_delete', array( $this, 'tutor_cart_delete' ) );
+			add_action( 'wp_ajax_tutor_course_from_cart', array( $this, 'tutor_course_from_cart' ) );
```
shewa12, line 97: I think this`wp_ajax_tutor_course_from_cart` should be `wp_ajax_tutor_delete_course_from_cart `

ecommerce/CartController.php
```diff
@@ -166,6 +166,40 @@ public function get_cart_by_id() {
...
+	}
+
+	/**
+	 * Get delete course from cart
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return void JSON response
+	 */
+	public function tutor_course_from_cart() {
```
shewa12, line 187: Same here method name to be clear

ecommerce/CartController.php
```diff
@@ -166,6 +166,40 @@ public function get_cart_by_id() {
...
+		if ( ! tutor_utils()->is_nonce_verified() ) {
+			$this->json_response( tutor_utils()->error_message( 'nonce' ), null, HttpHelper::STATUS_BAD_REQUEST );
```
shewa12, line 189: `$this->json_response( tutor_utils()->error_message( 'nonce' ), null, HttpHelper::STATUS_BAD_REQUEST );` I think this response is for API not for typical ajax request. Just use:
`wp_send_json_error( tutor_utils()->error_message( 'nonce' ) );`

shewa12 (changes requested):
Please check the request changes

ecommerce/CartController.php
sazedul-haque: Updated.

ecommerce/CartController.php
sazedul-haque: Updated.

ecommerce/CartController.php
sazedul-haque: Updated.

PR #917, opened 2024-07-26
🛠️ fix: Lesson preview switch field is linked with the lesson name field.

b-l-i-n-d:


assets/react/v3/entries/course-builder/components/modals/LessonModal.tsx
```diff
@@ -130,12 +134,21 @@ const LessonModal = ({
           prerequisites: lessonDetails.content_drip_settings?.course_prerequisites || [],
         },
       });
+
+      timeoutId = setTimeout(() => {
+        form.setFocus('title');
+      }, 0);
```
ahamed, line 140: What is the purpose of this?

assets/react/v3/entries/course-builder/components/modals/LessonModal.tsx
```diff
@@ -372,14 +385,17 @@ const LessonModal = ({
                 render={(controllerProps) => (
                   <FormSwitch
                     {...controllerProps}
+                    disabled={!isTutorPro || !isAddonEnabled('Tutor Course Preview')}
```
ahamed, line 388: Why passing `'Tutor Course Preview'` text as function param?

assets/react/v3/entries/course-builder/components/modals/LessonModal.tsx
b-l-i-n-d: This will make sure, form.setFocus will run after the whole form has been rendered. Without this the form.setFocus trigger does not show any change in the focus state on the front end. I tried triggering setFocus after resetting the whole form., it doesn't work.

assets/react/v3/entries/course-builder/components/modals/LessonModal.tsx
b-l-i-n-d: This function receives the addon name as `params`.  I have added `typescript` type `suggestion` support so the developer does not accidentally type wrong addon names.

assets/react/v3/entries/course-builder/components/modals/LessonModal.tsx
```diff
@@ -372,14 +385,17 @@ const LessonModal = ({
...
                         {__('Lesson Preview', 'tutor')}
-                        {!isAddonEnabled('Tutor Course Preview') && <SVGIcon name="crown" width={24} height={24} />}
+                        {!isTutorPro && !isAddonEnabled('Tutor Course Preview') && (
+                          <SVGIcon name="crown" width={24} height={24} />
+                        )}
                       </div>
                     }
                     helpText={
-                      isAddonEnabled('Tutor Course Preview')
+                      isTutorPro && isAddonEnabled('Tutor Course Preview')
```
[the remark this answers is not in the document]
b-l-i-n-d: Same here.

assets/react/v3/entries/course-builder/components/modals/LessonModal.tsx
ahamed: But shouldn't it be translatable? And it would be more readable if you put that text into some variable and pass the variable.

b-l-i-n-d: No, it shouldn't be translatable. It is just the addon name that is defined in `tutorConfig.addons_data`. I have defined an `enum Addons` that can be used as `Addons.addonName`. I will replace the string with `Addons.addonName`

ahamed (approved):

PR #923, opened 2024-07-26
Earnings & Commission Management

shewa12:
Store earning & commission on order creation. Example of **storing earnings**
```
$earnings = EarningController::instance();
$earnings->prepare_order_earnings( $order_id );
try {
	$earning_id = $earnings->remove_before_store_earnings();
	if ( $earning_id ) {
		do_action( 'tutor_ecommerce_after_earning_stored', $earning_id, $earnings->earning_data );
	}
} catch ( \Throwable $th ) {
	error_log( $th->getMessage() );
}
```


ecommerce/EarningController.php
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	/**
+	 * Instance of this class
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return EarningController object
+	 */
+	public static function instance() {
```
harunollyo, line 82: Instead of making individual class as singleton, we can use reusable `Singleton` class to make this class as singleton. PR https://github.com/themeum/tutor/pull/924

```php
class EarningController extends Singleton {

  protected function __construct() {
     global $wpdb;
     self::$instance->earning_table = $wpdb->prefix . 'tutor_earnings';
  }

```

ecommerce/EarningController.php
```diff
@@ -0,0 +1,300 @@
...
+	private $order_id;
+
+	/**
+	 * Keep earning data here
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array
+	 */
+	public $earning_data = array();
```
harunollyo, line 64: Make this property as `private` to ensure only this class method `prepare_order_earnings` can modify this.

To get safe prepared data, make a getter method
```php
public function get_earning_data() {
   return $this->earning_data;
}
```

Later we can use
```php
if ( $earning_id ) {
    do_action( 'tutor_ecommerce_after_earning_stored', $earning_id, $earnings->get_earning_data() );
}
```

harunollyo (changes requested):

ecommerce/EarningController.php
shewa12: `public $earning_data = array();` earning data need to keep public. `prepare_order_earnings` method is a special case. We may not need to prepare earning_data using order id always.

ecommerce/EarningController.php
```diff
@@ -0,0 +1,292 @@
...
+
+	/**
+	 * Instance of this class
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return EarningController object
+	 */
+	public static function get_instance() {
```
harunollyo, line 74: No need to override `get_instance` method and property `public static $instance = null;`

```php
protected function __construct() {
  global $wpdb;
  $this->earning_table = $wpdb->prefix . 'tutor_earnings';
}
```

harunollyo (changes requested):

harunollyo (approved):

PR #924, opened 2024-07-28
Singleton class added to make any class as singleton behaviour

harunollyo:
- New: `Singleton` class
- Unit test written
- Test result
image

### Run unit test
```
vendor/bin/phpunit --filter=SingletonClassTest
```

### How to use?

```php
class MySingletonClass extends Singleton {
   public function hello() {
       echo 'say hello';
   }
}

$obj = MySingletonClass::get_instance();
$obj->hello();
```


shewa12:
There is already a singleton mechanism exists

```
	function tutor_lms() {
		return \TUTOR\Tutor::instance();
	}
```

```
	public static function instance() {
		if ( is_null( self::$_instance ) ) {
			self::$_instance = new self();
		}
		return self::$_instance;
	}

```

harunollyo:
> There is already a singleton mechanism exists
> 
> ```
> 	function tutor_lms() {
> 		return \TUTOR\Tutor::instance();
> 	}
> ```
> 
> ```
> 	public static function instance() {
> 		if ( is_null( self::$_instance ) ) {
> 			self::$_instance = new self();
> 		}
> 		return self::$_instance;
> 	}
> ```



> There is already a singleton mechanism exists
> 
> ```
> 	function tutor_lms() {
> 		return \TUTOR\Tutor::instance();
> 	}
> ```
> 
> ```
> 	public static function instance() {
> 		if ( is_null( self::$_instance ) ) {
> 			self::$_instance = new self();
> 		}
> 		return self::$_instance;
> 	}
> ```

`Singleton` class is not tied to our tutor plugin instance. It's a separate parent class to make any class as singleton behaviour.

PR #963, opened 2024-08-12
Coupon create and update API integration added

sazedul-haque:


assets/react/v3/entries/coupon-details/components/coupon/CouponValidity.tsx
```diff
@@ -62,13 +65,15 @@ function CouponValidity() {
 						<BoxSubtitle css={styles.dateTimeTitle}>{__('Ends in', 'tutor')}</BoxSubtitle>
 						<div css={styles.dateTimeWrapper}>
 							<Controller
-								control={form.control}
 								name="end_date"
+								control={form.control}
+								rules={requiredRule()}
```
b-l-i-n-d, line 70: We need to add end-date validation so the user can not input the date before the start date.

b-l-i-n-d (approved):

PR #1002, opened 2024-08-26
Create openai client for making requests.

ahamed:


classes/Utils.php
```diff
@@ -1120,6 +1120,25 @@ public function check_nonce() {
 		}
 	}
 
+	/**
+	 * Check if the openai response has any error or not.
+	 * If there any error then send the error response, otherwise continue.
+	 *
+	 * @param array $response The openai response.
+	 * @return object
```
shewa12, line 1128: I think the return type would be mixed: object|null. `send_response` will just echo json data and die, in that case null will be returned.

composer.lock
```diff
@@ -8,26 +8,26 @@
...
                 "type": "zip",
-                "url": "https://api.github.com/repos/CarbonPHP/carbon-doctrine-types/zipball/99f76ffa36cce3b70a4a6abce41dba15ca2e84cb",
-                "reference": "99f76ffa36cce3b70a4a6abce41dba15ca2e84cb",
+                "url": "https://api.github.com/repos/CarbonPHP/carbon-doctrine-types/zipball/18ba5ddfec8976260ead6e866180bd5d2f71aa1d",
+                "reference": "18ba5ddfec8976260ead6e866180bd5d2f71aa1d",
                 "shasum": ""
             },
             "require": {
-                "php": "^7.4 || ^8.0"
+                "php": "^8.1"
```
shewa12, line 24: Need to revert the changes from composer.lock file

classes/Utils.php
```diff
@@ -1120,6 +1120,26 @@ public function check_nonce() {
...
+	/**
+	 * Check if the openai response has any error or not.
+	 * If there any error then send the error response, otherwise continue.
+	 *
+	 * @since   3.0.0
+	 *
+	 * @param array $response The openai response.
+	 * @return mixed
+	 */
+	public function check_openai_response( array $response ) {
```
harunollyo, line 1132: Why this helper here? It's related to `tutorai` we can move it to `tutorai/Helper.php`

harunollyo (changes requested):

classes/Utils.php
ahamed: I'll move it later. For now keep this as it is.

harunollyo (approved):

PR #1024, opened 2024-08-29
chore: Update usePortalPopover and modal component zIndex handling

b-l-i-n-d:
This commit updates `usePortalPopover` to fix scrolling and position and `Modal` component to handle the `zIndex` property. It replaces the usage of the `zIndex` constant from the `@Config/styles` module with the `configZIndex` constant. This change ensures consistent handling of the `zIndex` value throughout the component.

### DIFF
--- a/assets/react/v3/shared/components/modals/Modal.tsx
+++ b/assets/react/v3/shared/components/modals/Modal.tsx
@@ -1,4 +1,4 @@
-import { colorTokens, zIndex as configZIndex } from '@Config/styles';
+import { colorTokens, zIndex } from '@Config/styles';
 import { AnimatedDiv, AnimationType, useAnimation } from '@Hooks/useAnimation';
 import { nanoid, noop } from '@Utils/util';
 import { css } from '@emotion/react';
@@ -10,7 +10,7 @@ const styles = {
     background-color: ${colorTokens.background.modal};
     opacity: 0.7;
     inset: 0;
-    z-index: ${configZIndex.negative};
+    z-index: ${zIndex.negative};
 
     ${
       magicAi &&
@@ -22,7 +22,7 @@ const styles = {
     }
   `,
   container: css`
-    z-index: ${configZIndex.modal};
+    z-index: ${zIndex.modal};
     position: fixed;
     display: flex;
     justify-content: center;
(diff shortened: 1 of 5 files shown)


assets/react/v3/shared/components/modals/Modal.tsx
```diff
@@ -1,4 +1,4 @@
-import { colorTokens, zIndex } from '@Config/styles';
+import { colorTokens, zIndex as configZIndex } from '@Config/styles';
```
ahamed, line 1: Why? Could you please explain?

assets/react/v3/shared/components/modals/Modal.tsx
```diff
@@ -82,13 +82,21 @@ export const ModalProvider: React.FunctionComponent<{ children: ReactNode }> = (
   });
 
   const showModal = useCallback<ModalContextType['showModal']>(
-    ({ component, props, closeOnOutsideClick = false, isMagicAi = false }) => {
+    ({ component, props, closeOnOutsideClick = false, isMagicAi = false, zIndex = configZIndex.modal }) => {
```
ahamed, line 85: You can use `depthIndex` or only `depth` instead of `zIndex` here.

assets/react/v3/shared/hooks/usePortalPopover.tsx
```diff
@@ -173,21 +167,20 @@ interface PortalProps {
...
     return () => {
-      if (!hasModalOnStack) {
-        document.body.style.overflow = 'initial';
-      }
+      setTimeout(() => {
+        const hasPopoverOnStack = document.querySelectorAll('.tutor-portal-popover').length > 0;
+        if (!hasPopoverOnStack) {
+          document.body.style.overflow = 'initial';
+        }
+      }, 200); // this is the total of -> duration of the animation + threshold + 30ms
```
ahamed, line 181: Instead of commenting to recognising a magic value, put the value into a variable.
```ts
const ANIMATION_DURATION_WITH_THRESHHOLD = 200;
```

ahamed (changes requested):

assets/react/v3/shared/components/modals/Modal.tsx
b-l-i-n-d: As there was two same variables, but this would not be making any issue now as I have changed the other one to be `depthIndex` and kept it as `zIndex`.

ahamed (approved):

PR #1035, opened 2024-08-30
Update the empty state illustration

ahamed:


classes/Tutor.php
```diff
@@ -560,11 +560,6 @@ public function __construct() {
 		 */
 		new Ecommerce();
 
-		/**
-		 * Tutor AI generation controller
-		 */
-		new MagicAI();
```
sazedul-haque, line 566: @ahamed Vai, this seems irrelevant to this PR. Was it intentional or a accidental?

ahamed: Intentional.

sazedul-haque (approved):

PR #1061, opened 2024-09-07
Database Tables Create Only If Not Exists

shewa12:


classes/Upgrader.php
```diff
@@ -131,7 +131,13 @@ public function upgrade_to_2_6_0() {
 	 * @return void
 	 */
 	public function upgrade_to_3_0_0() {
-		Tutor::tutor_activate();
+		global $wpdb;
```
harunollyo, line 134: We can use `QueryHelper::table_exists( $table_name );`

PR #1083, opened 2024-09-12
Course builder - localized only required settings and added key to detect ChatGPT api key provided or not.

harunollyo:
https://app.clickup.com/t/86eq8ued6

### DIFF
--- a/classes/Options_V2.php
+++ b/classes/Options_V2.php
@@ -124,6 +124,20 @@ class Options_V2 {
 	}
 
 	/**
+	 * Get only list of options.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @param array $only only options.
+	 *
+	 * @return array
+	 */
+	public static function get_only( $only = array() ) {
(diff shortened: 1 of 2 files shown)


classes/Options_V2.php
```diff
@@ -123,6 +123,20 @@ private function get( $key = null, $default = false ) {
...
+	/**
+	 * Get only list of options.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @param array $only only options.
+	 *
+	 * @return array
+	 */
+	public static function get_only( $only = array() ) {
```
shewa12, line 135: `$only` argument should accept string as well for a single option

harunollyo: This is dedicated for fetching list of selected options. Mentioned in doc comment
``
Get only list of options.
``

We have already `get` and `tutor_utils()->get_option` for single option.

shewa12 (approved):

PR #1087, opened 2024-09-13
Codebase Refactor

shewa12:
1. **CURRENYC_SYMBOL** const replace with **CURRENCY_CODE** since storing currency code
2. `attach_product_with_course` Setting course type as paid & setting regular & sale price course price to make **Tutor** monetisation compatible in this method


classes/Course.php
```diff
@@ -2006,8 +2011,9 @@ public function attach_product_with_course( $post_ID, $post_data ) {
 
 		$course = get_post( $post_ID );
 
-		if ( 'wc' === $monetize_by ) {
+		update_post_meta( $post_ID, self::COURSE_PRICE_TYPE_META, 'paid' );
```
harunollyo, line 2014: Here we can use constant
```
self::PRICE_TYPE_PAID
```

classes/Course.php
```diff
@@ -2023,14 +2029,20 @@ public function attach_product_with_course( $post_ID, $post_data ) {
...
+				
 				$product_id  = self::create_wc_product( $course->post_title, $course_price, $sale_price, $attached_product_id );
+				$product_obj = wc_get_product( $product_id );
 				if ( $product_obj->is_type( 'subscription' ) ) {
 					update_post_meta( $attached_product_id, '_subscription_price', $course_price );
 				}
+
+				// Set course regular & sale price.
+				update_post_meta( $post_ID, self::COURSE_PRICE_META, $product_obj->get_regular_price() );
+				update_post_meta( $post_ID, self::COURSE_PRICE_TYPE_META, $product_obj->get_sale_price() );
```
harunollyo, line 2041: It should be `self::PRICE_TYPE_PAID`, `self::PRICE_TYPE_FREE` or other

harunollyo (changes requested):

harunollyo (approved):

PR #1198, opened 2024-10-07
Simplified the tax implementation, keep the admin tax settings only, revert the tax calculation functionalities.

ahamed:


ecommerce/Supports/Shop.php
```diff
@@ -0,0 +1,180 @@
...
+
+			$new_key        = $key . '_with_currency';
+			$data->$new_key = tutor_get_formatted_price( $data->$key );
+		}
+
+		return $data;
+	}
+
+	public static function calculate_taxable_amount( $price, $tax_rate ) {
+		$price    = $price ?? 0;
```
shewa12, line 64: I think Instead null operator we should use the ternary operator
```
$price    = $price ?? 0;
```
`$price = $price ? $price : 0`
For the null operator if only the value $price is null then it will set 0 if price is not null like empty string then $price will be set as empty string

ecommerce/Supports/Shop.php
```diff
@@ -0,0 +1,180 @@
...
+			}
+
+			$new_key        = $key . '_with_currency';
+			$data->$new_key = tutor_get_formatted_price( $data->$key );
+		}
+
+		return $data;
+	}
+
+	public static function calculate_taxable_amount( $price, $tax_rate ) {
```
shewa12, line 63: Doc comment is missing

ecommerce/Supports/Shop.php
```diff
@@ -0,0 +1,180 @@
...
+		$price    = $price ?? 0;
+		$tax_rate = $tax_rate ?? 0;
+
+		return ( $price * $tax_rate ) / 100;
+	}
+
+	public static function add_taxable_prices( $final_price, $tax_rate ) {
```
shewa12, line 70: Please add doc comment

ecommerce/Supports/Shop.php
```diff
@@ -0,0 +1,180 @@
...
+		if ( empty( $final_price ) || ! is_object( $final_price ) ) {
+			return $final_price;
+		}
+
+		$keys   = array( 'item_price', 'discounted_price' );
+		$suffix = '_with_tax';
+		$price  = clone $final_price;
```
shewa12, line 77: Any specific reason of using `clone` keyword?

models/CouponModel.php
```diff
@@ -508,13 +509,56 @@ public function get_coupon_by_id( $coupon_id ) {
...
+	}
+
+	/**
+	 * Get the list of the all automatic coupons.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return array
+	 */
+	public function get_automatic_coupons() {
```
shewa12, line 536: We can use the `get_coupons` method for this

models/CouponModel.php
```diff
@@ -508,13 +509,56 @@ public function get_coupon_by_id( $coupon_id ) {
 			return false;
 		}
 
+		return $this->process_coupon_data( $coupon_data );
+	}
+
+	public function get_coupon_by_code( $coupon_code ) {
```
shewa12, line 515: We can leverage the existing method 
```
$where = ['coupon_code' => $coupon_code];
public function get_coupon( array $where ) {}

```

models/CouponModel.php
```diff
@@ -508,13 +509,56 @@ public function get_coupon_by_id( $coupon_id ) {
...
+
+	private function process_coupon_data( $coupon_data ) {
 		$coupon_data->id                  = (int) $coupon_data->id;
 		$coupon_data->usage_limit_status  = ! empty( $coupon_data->total_usage_limit ) ? true : false;
 		$coupon_data->total_usage_limit   = (int) $coupon_data->total_usage_limit;
 		$coupon_data->is_one_use_per_user = ! empty( $coupon_data->per_user_usage_limit ) ? true : false;
 		$coupon_data->discount_amount     = (float) $coupon_data->discount_amount;
-		$coupon_data->created_by          = get_userdata( $coupon_data->created_by )->display_name;
-		$coupon_data->updated_by          = get_userdata( $coupon_data->updated_by )->display_name;
+		$coupon_data->created_by          = get_userdata( $coupon_data->created_by )->display_name ?? '';
```
shewa12, line 560: We can use the utility method like the following
`tutor_utils()->display_name( $user_id )`

shewa12 (changes requested):
Please check the requests for changes

harunollyo:
Country > State list will be sync later

harunollyo (approved):

shewa12 (approved):

PR #1275, opened 2024-10-29
Re-useable Helper added to get prices with tax info

harunollyo:
With
- Course list price column size updated to adjust price data
- Order type clause added to count order menu badge

Helper re-used in
https://github.com/themeum/tutor-pro/pull/728

### DIFF
--- a/classes/Utils.php
+++ b/classes/Utils.php
@@ -1228,22 +1228,7 @@ class Utils {
 			$regular_price = (float) get_post_meta( $course_id, Course::COURSE_PRICE_META, true );
 			$sale_price    = (float) get_post_meta( $course_id, Course::COURSE_SALE_PRICE_META, true );
 
-			$display_price       = $sale_price ? $sale_price : $regular_price;
-			$show_price_with_tax = Tax::show_price_with_tax();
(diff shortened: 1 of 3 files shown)


classes/Utils.php
```diff
@@ -1263,6 +1248,42 @@ public function get_raw_course_price( $course_id = 0 ) {
...
+	 */
+	public function get_prices_with_tax_info( $regular_price, $sale_price = null ) {
+
+		$display_price       = $sale_price ? $sale_price : $regular_price;
+		$show_price_with_tax = Tax::show_price_with_tax();
+		$user_logged_in      = is_user_logged_in();
+
+		$tax_amount = 0;
+		$tax_rate   = 0;
+		if ( $show_price_with_tax && is_numeric( $display_price ) && ! Tax::is_tax_included_in_price() ) {
```
shewa12, line 1269: `is_numeric( $display_price )`
Need to make sure there is no comma (**10,100.10**) in the price otherwise is_numeric will not work

harunollyo: We are only accepting `int|flat` value in param without formatted value.

shewa12 (approved):

PR #1316, opened 2024-11-06
Fix : When user buy multiple products, earning shows based on the last product(Report/Annalytics)

Anindra123:


classes/Earnings.php
```diff
@@ -202,7 +202,9 @@ public function store_earnings() {
 		}
 
 		try {
-			$inserted_id = QueryHelper::insert( $this->earning_table, $this->earning_data );
+			foreach ( $this->earning_data as $earning_data ) {
+				$inserted_id = QueryHelper::insert( $this->earning_table, $earning_data );
```
shewa12, line 206: Declare `$insert_id` up above the foreach loop

PR #1327, opened 2024-11-08
Order history and refund modal updates 

sazedul-haque:


assets/react/v3/entries/order-details/components/modals/RefundModal.tsx
```diff
@@ -56,7 +57,7 @@ function RefundModal({ title, closeModal, actions, available_amount, order_id, o
               rules={{
                 ...requiredRule(),
                 validate: (value) => {
-                  if (value === 0) {
+                  if (value == 0) {
```
b-l-i-n-d, line 60: We can use `Number(value)`

b-l-i-n-d (approved):

PR #1330, opened 2024-11-08
Fix - coupon minimum purchase requirement not work for subscription checkout

harunollyo:


models/CouponModel.php
```diff
@@ -845,19 +845,28 @@ public function is_coupon_applicable( object $coupon, int $object_id ): bool {
...
+				$course_price = tutor_utils()->get_raw_course_price( $item_id );
+				if ( OrderModel::TYPE_SINGLE_ORDER !== $order_type ) {
+					$plan_info = apply_filters( 'tutor_checkout_plan_info', null, $item_id );
+					if ( $plan_info ) {
+						$course_price->regular_price = $plan_info->regular_price;
+						$course_price->sale_price    = $plan_info->in_sale_price ? $plan_info->sale_price : 0;
+					}
+				}
+
 				$total_price += $course_price->sale_price ? $course_price->sale_price : $course_price->regular_price;
```
shewa12, line 870: Instead of this 
`$total_price += $course_price->sale_price ? $course_price->sale_price : $course_price->regular_price;`

We can use existing method
`$total_price += OrderModel::get_item_sellable_price( $course_price->regular_price, $course->sale_price );`

PR #1380, opened 2024-11-19
🛠️fix: Payemnt Methods Validation Added

b-l-i-n-d:


assets/react/v3/shared/components/fields/FormInput.tsx
```diff
@@ -253,6 +253,11 @@ const styles = {
     button {
       padding: ${spacing[8]};
       border-radius: ${borderRadius[2]};
+
+      :focus-visible {
```
sazedul-haque, line 257: This should already be available from Buttom atom.

sazedul-haque (changes requested):

sazedul-haque (approved):

PR #1422, opened 2024-11-25
Feat: added edit with divi builder

Anindra123:


classes/Utils.php
```diff
@@ -10309,6 +10323,10 @@ public function get_editor_used( $post_id ) {
 			$name = 'droip';
 		}
 
+		if ( 'on' === get_post_meta( $post_id, '_et_pb_use_builder', true ) ) {
```
harunollyo, line 10326: 1. Need to implement unlink divi by deleting this meta `_et_pb_use_builder` in `Course.php > ajax_unlink_page_builder`
2. Move this check before `droip`

harunollyo (changes requested):

PR #1483, opened 2024-12-24
New settings added for instructors

sazedul-haque:


b-l-i-n-d (approved):

classes/Course.php
```diff
@@ -999,6 +999,11 @@ public function ajax_update_course() {
 			}
 		}
 
+		$is_error = apply_filters( 'tutor_is_error_before_course_update', false, $params );
+		if ( is_wp_error( $is_error ) ) {
+			$this->json_response( $is_error->get_error_message(), null, HttpHelper::STATUS_INTERNAL_SERVER_ERROR );
```
harunollyo, line 1004: Use bad request instead of internal error.
```php
$this->response_bad_request( $is_error->get_error_message() );
```

harunollyo (changes requested):

harunollyo (approved):

PR #1519, opened 2025-01-13
Refactor the ajax_course_list as per new design & API request

shewa12:


classes/Course.php
```diff
@@ -794,14 +794,28 @@ public function ajax_create_new_draft_course() {
 	 *
 	 * @since 3.0.0
 	 *
+	 * @since 3.2.0
```
harunollyo, line 797: Make the since changelog inline

classes/Course.php
```diff
@@ -812,28 +826,28 @@ function( $id ) {
...
+				$response['results'][] = self::get_mini_info( $course );
 			}
-
-			$items[] = $tmp;
 		}
 
 		$this->json_response(
-			__( 'Course list fetched successfully', 'tutor' ),
-			$items
+			__( 'Course list retrieved successfully!', 'tutor-pro' ),
```
harunollyo, line 849: Invalid text domain

harunollyo (changes requested):

PR #1532, opened 2025-01-16
Fix - empty name on order history

harunollyo:


templates/dashboard/purchase_history.php
```diff
@@ -141,15 +141,15 @@
 									$items = ( new OrderModel() )->get_order_items_by_id( $order->id );
 									foreach ( $items as $item ) {
 										$course_id    = $item->id;
-										$object_title = '';
+										$object_title = get_the_title( $course_id );
```
shewa12, line 144: Since we are calling object title then why the variable name is **$course_id** it should be **$object_id**

harunollyo: Because the immediate $course_id is for  single course, bundle order

shewa12: If there is nothing other than a course, we can use $course_title. Mixed variable names like $course_id & $object_title seems confusing.

harunollyo: Already updated

shewa12 (approved):

PR #1548, opened 2025-01-22
Refactor recurring limit to use translation function for consistency

b-l-i-n-d:


sazedul-haque:
I think you missed one string `SubscriptionItem.tsx`

sazedul-haque (approved):

PR #1549, opened 2025-01-22
🛠️Fix: Translation fixes

b-l-i-n-d:


assets/react/v3/shared/components/subscription/OfferSalePrice.tsx
```diff
@@ -58,7 +58,7 @@ export function OfferSalePrice({ index }: { index: number }) {
               <FormInputWithContent
                 {...props}
                 type="number"
-                label="Sale Price"
+                label={__('Sale price', 'tutor')}
```
sazedul-haque, line 61: We are using `title-case` for label in tutor. It should be `Sale Price`

sazedul-haque (approved):

PR #1568, opened 2025-01-30
Update: Password protected course and bundle enrollment fix and show error on form submission

Anindra123:


classes/Course.php
```diff
@@ -268,9 +268,27 @@ public function __construct( $register_hooks = true ) {
...
 
+	/**
+	 * Handle password protected course and bundle form submission.
+	 *
+	 * @sine 3.2.1
+	 *
+	 * @return void
+	 */
+	public function handle_password_submit() {
+		if ( Input::has( 'post_password' ) ) {
```
harunollyo, line 283: Need to add check for `course_id` also. Otherwise this will impact on blog post
```php
Input::has( 'course_id' );
```

harunollyo (changes requested):

harunollyo (approved):

PR #1583, opened 2025-02-10
✨ Feat: ErrorBoundary refactored for Production

b-l-i-n-d:


sazedul-haque (changes requested):
A better approach would be creating two different files and import dynamically. This will reduce production bundle size and produce cleaner code. We can create an ErrorBoundaryWrapper like the example bellow which will server the desired component.

```
import React, { ReactNode } from "react";

let ErrorBoundary: React.ComponentType<{ children: ReactNode }>;

if (process.env.NODE_ENV === "development") {
  ErrorBoundary = require("@TutorShared/components/ErrorBoundaryDev").default;
} else {
  ErrorBoundary = require("@TutorShared/components/ErrorBoundaryProd").default;
}

export default ErrorBoundary;
```

sazedul-haque (approved):

PR #1611, opened 2025-02-25
Fix(🛠️): Double payment issue for 2checkout

sanjana4khan:


models/OrderModel.php
```diff
@@ -1689,21 +1689,60 @@ public static function is_manual_payment( $method_name ) {
...
+	 *
+	 * @since 3.3.0
+	 *
+	 * @param object $order The order object containing order details.
+	 *
+	 * @return bool Returns true if the order is expired or expiry time is not set, otherwise false.
+	 */
+	private static function check_expiry_time( $order ) {
+
+		$current_time = DateTimeHelper::now()->to_timestamp();
```
shewa12, line 1732: Instead of `DateTimeHelper::now()->to_timestamp()` we can use `time()`;

models/OrderModel.php
```diff
@@ -1689,21 +1689,60 @@ public static function is_manual_payment( $method_name ) {
...
+		$meta_key     = "tutor_order_id_$order->id";
+		$meta_value   = get_user_meta( get_current_user_id(), $meta_key, false );
+		$expiry_time  = ! empty( $meta_value ) ? json_decode( $meta_value[0] )->expiry_time : null;
+
+		if ( ! is_null( $expiry_time ) && $expiry_time < $current_time ) {
```
shewa12, line 1737: Instead of `! is_null( $expiry_time )`, just use `!$expiry_time`. It seems logic operand should be OR.

models/OrderModel.php
```diff
@@ -1689,21 +1689,60 @@ public static function is_manual_payment( $method_name ) {
...
+			delete_user_meta( get_current_user_id(), $meta_key );
```
shewa12, line 1738: Instead of using `get_current_user_id ` function repeatedly, keep it in a variable.

models/OrderModel.php
```diff
@@ -1689,21 +1689,60 @@ public static function is_manual_payment( $method_name ) {
...
+			return true;
+		}
+
+		if ( ! is_null( $expiry_time ) && $expiry_time > $current_time ) {
```
shewa12, line 1742: Use `!$expiry_time` condition instead of null checking

shewa12 (changes requested):
Please do the request changes.

shewa12 (approved):

PR #1631, opened 2025-03-05
Fix: Co-instructor cannot see course on backend

Anindra123:


classes/Course_List.php
```diff
@@ -263,7 +263,7 @@ protected static function count_course( string $status, $category_slug = '', $co
 			);
 		}
 
-		$the_query = new \WP_Query( $args );
+		$the_query = new \WP_Query( apply_filters( 'tutor_admin_course_list', $args, $user_id, $status ) );
```
shewa12, line 266: Create a reusable function for the course wp_query to prevent duplicate code. Write now we have the following script
`new \WP_Query( apply_filters( 'tutor_admin_course_list', $args, $user_id, $status ) );` in 2 places in future it may increase.

shewa12 (approved):

PR #1645, opened 2025-03-10
Fix: Assignment expiration based on enrollment date

Anindra123:


classes/Utils.php
```diff
@@ -8171,9 +8171,14 @@ public function has_enrolled_content_access( $content, $object_id = 0, $user_id
 	 * @param mixed $format format.
 	 * @param mixed $fallback fallback.
 	 *
+	 * @since 3.4.0
```
harunollyo, line 8174: Add this after first since tag and since comment meta should be inline

templates/dashboard/assignments/submitted.php
```diff
@@ -120,6 +125,12 @@
 									</div>
 								</div>
 							</td>
+							<td>
+								<?php echo wp_kses_post( tutor_utils()->convert_date_into_wp_timezone( $assignment->comment_date_gmt , 'j M, Y,<\b\r>h:i a' ) ); ?>
+							</td>
+							<td>
+								<?php echo esc_html( DateTimeHelper::get_gmt_to_user_timezone_date( $deadline_date ), 'j M, Y, h:i a' ); ?>
```
harunollyo, line 132: No need 2nd param. It should be time format which admin set.

harunollyo (changes requested):

harunollyo (approved):

PR #1686, opened 2025-03-24
Fix: Tax not calculating when manual payment by student

Anindra123:


models/OrderModel.php
```diff
@@ -160,6 +161,35 @@ public function get_table_name() {
 		return $this->table_name;
 	}
 
+	/**
+	 * Obtain order tax data when student pays.
```
harunollyo, line 165: Get recalculated order tax data.

harunollyo (changes requested):

harunollyo (approved):

PR #1697, opened 2025-04-09
🛠️ Refactor supportRandomize question types by removing 'ordering'

b-l-i-n-d:
Eliminate 'ordering' from the supportRandomize question types to streamline the available options.


sazedul-haque (changes requested):
Only changing will not work. currently if this toggle is disabled, options will not display randomly.

sazedul-haque (approved):

PR #1718, opened 2025-04-16
Fix: Assignment fatal error due to date time format

Anindra123:


classes/Utils.php
```diff
@@ -10,6 +10,7 @@
 
 namespace TUTOR;
 
+use DateTime;
```
harunollyo, line 13: Unused `DateTime` import

classes/Utils.php
```diff
@@ -8199,7 +8197,7 @@ public function get_assignment_deadline_date( $assignment_id, $format = null, $f
 		$date = date_create( $deadline_date );
 		date_add( $date, date_interval_create_from_date_string( $value . ' ' . $time ) );
 
-		return date_format( $date, $format );
+		return date_format( $date, 'Y-m-d H:i:s' );
```
harunollyo, line 8200: Use `DateTimeHelper::FORMAT_MYSQL` constant

harunollyo (changes requested):

harunollyo (approved):

PR #1726, opened 2025-04-18
Fix: Prevent scheduled course preview of other instructor

Anindra123:


shewa12 (changes requested):
pls check the request changes

PR #1743, opened 2025-04-22
💳 Add `net-payment` condition to handle PaymentActionButton

b-l-i-n-d:
Implement net payment to handle the PaymentActionButton component and update relevant imports. This change ensures that the button does not render when the net payment is zero or less.

Single Purchase using 100% coupon
image

Subscription using 100% coupon
image

Membership with trial
image

### DIFF
--- a/assets/react/v3/entries/order-details/components/order/Payment.tsx
+++ b/assets/react/v3/entries/order-details/components/order/Payment.tsx
@@ -28,9 +28,9 @@ function PaymentActionButton({
   order: Order;
   onClick: (buttonType: 'refund' | 'mark-as-paid') => void;
 }) {
-  const { payment_status, total_price, net_payment } = order || {};
+  const { payment_status, total_price } = order || {};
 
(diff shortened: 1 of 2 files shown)


assets/react/v3/entries/order-details/components/order/Payment.tsx
```diff
@@ -1,30 +1,40 @@
...
   onClick,
 }: {
-  status: PaymentStatus;
+  order: Order;
   onClick: (buttonType: 'refund' | 'mark-as-paid') => void;
 }) {
-  switch (status) {
+  const { payment_status, net_payment } = order || {};
+
+  if (net_payment <= 0) {
```
sazedul-haque, line 33: What if the payment_status is `unpaid`. In that case we need to show the Mark as paid button

sazedul-haque (changes requested):

sazedul-haque (approved):

PR #1748, opened 2025-04-23
Fixed The _load_textdomain_just_in_time Notice Error

shewa12:
**Problem:**
WordPress _load_textdomain_just_in_time load the translation after the initialization of WP. When the translation functions like __, esc_html_e get called before loading the translation functions it throws error.

**Solution:**
Calling the translation functions after loading WP solved the issue. Used the `init` action hook. Thanks to the WP hooks. Below is an example of solution:

**Previous Code:**
`$this->page_title = __( 'Quiz Attempts', 'tutor' );`

**Current Code:**
```
		add_action(
			'init',
			function() {
				$this->page_title = __( 'Quiz Attempts', 'tutor' );
			}
		);
```

### DIFF
--- a/classes/Announcements.php
+++ b/classes/Announcements.php
@@ -49,7 +49,13 @@ class Announcements {
 	 * @return void
 	 */
 	public function __construct() {
-		$this->page_title = __( 'Announcements', 'tutor' );
+		add_action(
+			'init',
+			function() {
+				$this->page_title = __( 'Announcements', 'tutor' );
+			}
+		);
+
 		/**
 		 * Handle bulk action
 		 *
--- a/classes/Course_Widget.php
+++ b/classes/Course_Widget.php
@@ -32,8 +32,8 @@ class Course_Widget extends \WP_Widget {
 	public function __construct() {
 		parent::__construct(
 			'tutor_course_widget', // Base ID.
-			'Tutor Course',
-			array( 'description' => 'Display courses wherever widget support is available.' )
+			_x( 'Tutor Course', 'widget title', 'tutor' ),
+			array( 'description' => _x( 'Display courses wherever widget support is available.', 'widget description', 'tutor' ) )
 		);
 	}
 
--- a/classes/Course_Widget.php
+++ b/classes/Course_Widget.php
@@ -32,8 +32,8 @@ class Course_Widget extends \WP_Widget {
 	public function __construct() {
 		parent::__construct(
 			'tutor_course_widget', // Base ID.
-			esc_html__( 'Tutor Course', 'tutor' ), // Name.
-			array( 'description' => esc_html__( 'Display courses wherever widget support is available.', 'tutor' ) ) // Args.
+			'Tutor Course',
+			array( 'description' => 'Display courses wherever widget support is available.' )
 		);
 	}
 
--- a/classes/Instructors_List.php
+++ b/classes/Instructors_List.php
@@ -59,7 +59,12 @@ class Instructors_List {
 	 * @return void
 	 */
 	public function __construct() {
-		$this->page_title = __( 'Instructor', 'tutor' );
+		add_action(
+			'init',
+			function() {
+				$this->page_title = __( 'Instructor', 'tutor' );
+			}
+		);
 
 		/**
 		 * Handle bulk action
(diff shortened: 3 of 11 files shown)


classes/Announcements.php
```diff
@@ -49,7 +49,13 @@ class Announcements {
 	 * @return void
 	 */
 	public function __construct() {
-		$this->page_title = __( 'Announcements', 'tutor' );
+		add_action(
```
harunollyo, line 52: Instead of 6 line of code, we can write it in on-liner
```php
add_action( 'init', fn() => $this->page_title = __( 'Announcements', 'tutor' ) );
```

classes/Course_Widget.php
```diff
@@ -32,8 +32,8 @@ class Course_Widget extends \WP_Widget {
 	public function __construct() {
 		parent::__construct(
 			'tutor_course_widget', // Base ID.
-			esc_html__( 'Tutor Course', 'tutor' ), // Name.
-			array( 'description' => esc_html__( 'Display courses wherever widget support is available.', 'tutor' ) ) // Args.
+			'Tutor Course',
```
harunollyo, line 35: How this will be translated?

classes/Course_Widget.php
```diff
@@ -32,8 +32,8 @@ class Course_Widget extends \WP_Widget {
...
+			array( 'description' => 'Display courses wherever widget support is available.' )
```
harunollyo, line 36: Here also

classes/Instructors_List.php
```diff
@@ -59,7 +59,12 @@ class Instructors_List {
 	 * @return void
 	 */
 	public function __construct() {
-		$this->page_title = __( 'Instructor', 'tutor' );
+		add_action(
```
harunollyo, line 62: Instead of 6 line of code, we can write it in on-liner

classes/Quiz_Attempts_List.php
```diff
@@ -55,8 +55,13 @@ class Quiz_Attempts_List {
 	 * @param boolean $register_hook should register hook or not.
 	 */
 	public function __construct( $register_hook = true ) {
+		add_action(
```
harunollyo, line 58: Same here

classes/Quiz_Attempts_List.php
```diff
@@ -143,7 +148,7 @@ public function get_quiz_attempts_stat() {
 				$select_stmt = "SELECT COUNT( DISTINCT attempt_id)
 								FROM {$wpdb->prefix}tutor_quiz_attempts quiz_attempts
 								INNER JOIN {$wpdb->posts} quiz ON quiz_attempts.quiz_id = quiz.ID
-								-- INNER JOIN {$wpdb->prefix}tutor_quiz_attempt_answers AS ans ON quiz_attempts.attempt_id = ans.quiz_attempt_id";
+								INNER JOIN {$wpdb->prefix}tutor_quiz_attempt_answers AS ans ON quiz_attempts.attempt_id = ans.quiz_attempt_id";
```
harunollyo, line 151: Is this intentional or by mistake?

classes/Students_List.php
```diff
@@ -61,7 +61,13 @@ class Students_List {
 	 * @since 2.0.0
 	 */
 	public function __construct() {
-		$this->page_title = __( 'Students', 'tutor' );
+		add_action(
```
harunollyo, line 64: Instead of 6 line of code, we can write it in on-liner

classes/Withdraw_Requests_List.php
```diff
@@ -44,7 +44,13 @@ class Withdraw_Requests_List {
 	 * @since 1.0.0
 	 */
 	public function __construct() {
-		$this->page_title = __( 'Withdraw Request', 'tutor' );
+		add_action(
```
harunollyo, line 47: same here

ecommerce/CouponController.php
```diff
@@ -89,7 +89,13 @@ class CouponController extends BaseController {
 	 * @return void
 	 */
 	public function __construct( $register_hooks = true ) {
-		$this->page_title = __( 'Coupons', 'tutor' );
+		add_action(
```
harunollyo, line 92: Same here

ecommerce/OrderController.php
```diff
@@ -83,8 +83,14 @@ class OrderController {
 	 * @return void
 	 */
 	public function __construct( $register_hooks = true ) {
-		$this->page_title = __( 'Orders', 'tutor' );
-		$this->model      = new OrderModel();
+		add_action(
```
harunollyo, line 86: same here

harunollyo (changes requested):

classes/Announcements.php
shewa12: We can not do this as per this warning: **Assignments must be the first block of code on a line**

PR #1797, opened 2025-05-09
Multiple issue fixes

harunollyo:
- Fix: automatic WC product creating on pricing model update
- Fix: free course get enrolled when a paid course is purchased
- Fix: typo


classes/Course.php
```diff
@@ -2224,16 +2221,20 @@ public function tutor_add_gutenberg_author( $data, $postarr ) {
 	 * @return void
 	 */
 	public function attach_product_with_course( $post_ID, $post_data ) {
+		/**
+		 * For native monetization, just return
+		 * No need to attach anything.
+		 */
+		if ( tutor_utils()->is_monetize_by_tutor() ) {
```
shewa12, line 2228: This method adds a redundant database call. First checking
`tutor_utils()->is_monetize_by_tutor()` database called
Then again: 
`$monetize_by = tutor_utils()->get_option( 'monetize_by' );` database called for the same thing

It's better to reinstate the code snippet

```
		$monetize_by = tutor_utils()->get_option( 'monetize_by' );
		$product_id  = Input::post( '_tutor_course_product_id', 0, Input::TYPE_INT );

		if ( Ecommerce::MONETIZE_BY === $monetize_by ) {
			return;
		}
```

classes/Course.php
```diff
@@ -2265,16 +2266,15 @@ public function attach_product_with_course( $post_ID, $post_data ) {
 
 		if ( 'wc' === $monetize_by ) {
 
-			$is_update = ( $attached_product_id && wc_get_product( $attached_product_id ) ) ? true : false;
+			$is_update = ( $product_id && wc_get_product( $product_id ) ) ? true : false;
```
shewa12, line 2269: Here we are calling `wc_get_product` fn and on the line no: **2275** again calling `wc_get_product`
Which results in unoptimized code. Instead, we can keep it in a variable, then use it where required.

shewa12 (changes requested):
Please check the request changes

classes/Course.php
harunollyo: We can't do that because the price needs to be kept up to date during creation or updates.

PR #1838, opened 2025-05-22
Fixed: PHP error call to a member function on bool

shewa12:


models/CourseModel.php
```diff
@@ -87,11 +87,15 @@ public static function get_status_list() {
...
+	 * @since 3.6.0 $post_type param added
+	 *
+	 * @param string $status Post status.
+	 * @param string $post_type Post type.
+	 *
 	 * @return int
 	 */
-	public static function count( $status = self::STATUS_PUBLISH ) {
-		$count_obj = wp_count_posts( self::POST_TYPE );
+	public static function count( $status = self::STATUS_PUBLISH, $post_type = self::POST_TYPE ) {
```
harunollyo, line 97: Why this file change here?

harunollyo (changes requested):

PR #1852, opened 2025-05-27
🆕 feat: add delete functionality for import/export history

b-l-i-n-d:


assets/react/v3/entries/import-export/components/History.tsx
```diff
@@ -113,6 +128,34 @@ const History = () => {
         return <div css={styles.historyTitle}>{item.option_value.created_at}</div>;
       },
     },
+    {
+      Cell: (item, index) => {
+        return (
+          <div css={styles.action}>
```
sazedul-haque, line 134: Using dropdown only for 1 item is not ideal. We can use the delete button directly.

shewa12 (approved):

PR #1861, opened 2025-05-28
Refactored the get_course_reviews method to get reviews by course or comment ID

shewa12:
`is_course_object` new argument added to build the where clause. If `false` is provided, it will be used as a comment.

### DIFF
--- a/classes/Utils.php
+++ b/classes/Utils.php
@@ -4063,24 +4063,30 @@ class Utils {
 	 *
 	 * @since 1.0.0
 	 *
+	 * @since 3.6.0
+	 *
+	 * Fetching data by context, either course or comment
+	 *
 	 * @param int   $object id Course/Comment id.
 	 * @param int   $start offset.
 	 * @param int   $limit limit.
 	 * @param bool  $count_only count only.
 	 * @param array $status_in status list.
 	 * @param int   $include_user_id include user id.
+	 * @param int   $is_course_object Whether fetching by course or not.
 	 *
 	 * @return array|null|object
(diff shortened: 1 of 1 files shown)


classes/Utils.php
```diff
@@ -4072,10 +4072,17 @@ public function get_tutor_user( $user_id ) {
...
 	 * @return array|null|object
 	 */
-	public function get_course_reviews( $course_id = 0, $start = 0, $limit = 10, $count_only = false, $status_in = array( 'approved' ), $include_user_id = 0 ) {
-		$course_id = $this->get_post_id( $course_id );
+	public function get_course_reviews( $object_id = 0, $start = 0, $limit = 10, $count_only = false, $status_in = array( 'approved' ), $include_user_id = 0, $is_course_object = true  ) {
 		global $wpdb;
 
+		$where_clause = '_reviews.comment_post_ID = %d';
+
+		$object_id = (int) $this->get_post_id( $object_id );
```
harunollyo, line 4080: We can query post only when it's needed.
```php
$object_id = $is_course_object ? $this->get_post_id( $object_id ) : $object_id;
```

classes/Utils.php
```diff
@@ -4063,7 +4063,7 @@ public function get_tutor_user( $user_id ) {
 	 *
 	 * @since 1.0.0
```
harunollyo, line 4064: @since 3.6.0 param $is_course_object is added and update param $course_id to $object_id

harunollyo (changes requested):

PR #1944, opened 2025-07-09
✨ Integrate Content Bank functionality in QuestionList and TopicFooter components

b-l-i-n-d:
Add Content Bank integration to the QuestionList and TopicFooter components, enabling users to access and add content from the Content Bank based on their subscription status.


assets/react/v3/entries/course-builder/components/curriculum/QuestionList.tsx
```diff
@@ -240,6 +240,7 @@ const QuestionList = ({ isEditing }: { isEditing: boolean }) => {
       return {
         ...question,
         _data_status: QuizDataStatus.UPDATE,
+        is_cb_question: true,
```
sazedul-haque, line 243: Also need to add this to the converter.

sazedul-haque (changes requested):

PR #1960, opened 2025-07-14
✨ Enhance Select component and FocusTrap functionality

b-l-i-n-d:
Add transition effects and disabled state styles to the Select component. Implement a blur effect for previously active elements in FocusTrap and set tabIndex on the Logo button for improved accessibility.


sazedul-haque (approved):
Change `base` to 3.7.0. `content-bank` PR is merged.

PR #1988, opened 2025-07-23
🔧 Fix value type handling in FormInputWithContent and currency formatting

b-l-i-n-d:
Ensure correct value types are handled in `FormInputWithContent` and improve currency formatting by explicitly converting prices to numbers.


assets/react/v3/shared/components/fields/FormInputWithContent.tsx
```diff
@@ -85,7 +85,7 @@ const FormInputWithContent = ({
                     ? e.target.value.replace(/[^0-9.]/g, '').replace(/(\..*)\./g, '$1')
                     : e.target.value;
 
-                field.onChange(type === 'number' ? Number(value) : value);
+                field.onChange(value);
```
sazedul-haque, line 88: We should also remove `number` from return type. Also needs to test properly in every projects.

sazedul-haque (approved):

PR #1991, opened 2025-07-23
Refactor: Use reusable helper to display attempt result consistently across all locations

harunollyo:


models/QuizModel.php
```diff
@@ -1140,4 +1144,85 @@ public static function get_quiz_details( $quiz_id ) {
...
+		/**
+		 * Backward compatibility for getting a attempt result
+		 *
+		 * @since 3.7.1
+		 */
+		$attempt_info = maybe_unserialize( $attempt->attempt_info );
+		$answers_map  = self::get_quiz_answers_by_attempt_id( $attempt_id, true );
+
+		$earned_percentage = self::calculate_attempt_earned_percentage( $attempt );
+		$passing_grade     = (int) $attempt_info['passing_grade'] ?? 0;
```
shewa12, line 1210: `$passing_grade  = (int) $attempt_info['passing_grade'] ?? 0;`
I believe it is passing mark, maybe the name has been chosen based on how we are storing it in the DB, in this case, we can leave a comment to clear confusion.

shewa12 (approved):

PR #2001, opened 2025-07-25
Hook: New filter hook to check if user can complete course

shewa12:
This hook added to prevent the user to complete the course. If the hook return `wp_error` then the error message will be shown as toast message.

### DIFF
--- a/classes/Course.php
+++ b/classes/Course.php
@@ -2095,6 +2095,9 @@ class Course extends Tutor_Base {
 	 * Mark complete completed
 	 *
 	 * @since 1.0.0
+	 *
+	 * @since 3.7.1 Filter hook: tutor_user_can_complete_course added
+	 *
 	 * @return void
 	 */
 	public function mark_course_complete() {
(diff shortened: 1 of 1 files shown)


classes/Course.php
```diff
@@ -2114,15 +2119,22 @@ public function mark_course_complete() {
...
-		$permalink = get_the_permalink( $course_id );
-
-		// Set temporary identifier to show review pop up.
-		self::set_review_popup_data( $user_id, $course_id, $permalink );
+		/**
+		 * Filter hook provided to restrict course completion. This is useful
+		 * for specific cases like prerequisites. WP_Error should be returned
+		 * from the filter value to prevent the completion.
+		 */
+		$can_complete = apply_filters( 'tutor_user_can_complete_course', null, $user_id, $course_id );
```
harunollyo: Consider changing the default value from null to `true` to better align with the purpose of the `tutor_user_can_complete_course filter` and the variable name `$can_complete`. Using true as the default makes the logic clearer (allowed by default) and avoids unnecessary null checks.

classes/Course.php
```diff
@@ -2114,15 +2119,22 @@ public function mark_course_complete() {
...
+		if ( ! is_null( $can_complete ) && is_wp_error( $can_complete ) ) {
```
harunollyo: Here we can only check
```php
if ( is_wp_error( $can_complete ) ) {
  tutor_utils()->redirect_to( $permalink, $can_complete->get_error_message(), 'error' );
}
```

harunollyo (changes requested):

PR #2022, opened 2025-08-04
Fix: security issues

harunollyo:


classes/Utils.php
```diff
@@ -10806,4 +10809,17 @@ public function get_script_locale_data( string $filename, string $locale = 'en_U
...
+	/**
+	 * Get valid sort order.
+	 *
+	 * @since 3.7.1
+	 *
+	 * @param string $order order.
+	 *
+	 * @return string
+	 */
+	public function get_valid_sort_order( $order ) {
```
shewa12: This method should be in the QueryHelper class.

PR #2023, opened 2025-08-05
✨ Add Storybook integration and enhance component documentation

b-l-i-n-d:
Integrate Storybook for improved component development and documentation. Add stories for new components and enhance existing ones with better accessibility and type safety. Update configurations and dependencies for optimal performance.

Please run `npm i && npm run storybook`


assets/react/v3/shared/config/config.ts
```diff
@@ -1,10 +1,104 @@
-export const tutorConfig = window._tutorobject;
+const defaultTutorConfig = {
```
sazedul-haque, line 1: Why we need this? If `_tutorobject` not found it should `null` or `undefined`.

sazedul-haque (approved):

assets/react/v3/shared/config/config.ts
b-l-i-n-d: We need to access `tutorConfig` in many of our `atoms`, `molecules`, or `components`. Since `Storybook` does not interact with `WordPress`, `tutorConfig` remains undefined. I added a `defaultConfig` because if there is no `tutorConfig`, it will always fall back to default values, ensuring the code does not break.

PR #2024, opened 2025-08-06
Fix payment method install and activation logic

sazedul-haque:
Updates PaymentGatewayItem to correctly set is_installed and is_plugin_active when installing payment methods, ensuring new methods are appended only if not existing. Adjusts PaymentItem badge display logic to show 'Plugin Not Activated' only for non-manual methods.

### DIFF
--- a/assets/react/v3/entries/payment-settings/components/PaymentGatewayItem.tsx
+++ b/assets/react/v3/entries/payment-settings/components/PaymentGatewayItem.tsx
@@ -30,13 +30,13 @@ const PaymentGatewayItem = ({ data, onInstallSuccess, form }: PaymentGatewayItem
       let isPaymentExisting = false;
 
       // Mark as is_installed and is_plugin_active if it already exists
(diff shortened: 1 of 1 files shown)


assets/react/v3/entries/payment-settings/components/PaymentGatewayItem.tsx
```diff
@@ -27,27 +27,28 @@ const PaymentGatewayItem = ({ data, onInstallSuccess, form }: PaymentGatewayItem
       onInstallSuccess();
 
       const existingPaymentMethods = form.getValues('payment_methods') ?? [];
-      // Make is_installed false if already exist
+      let isPaymentExisting = false;
+
+      // Mark as is_installed and is_plugin_active if it already exists
       existingPaymentMethods.forEach((method) => {
```
b-l-i-n-d: Use `for` or `for...of` for performance improvements.

PR #2048, opened 2025-08-15
feat(course-filter): add toggle fields for course archive filter

nur-alam:
Add toggle_fields to course_archive_filter option to control visibility of related fields Refactor showHideToggleChildren function for better element selection and border handling

### DIFF
--- a/classes/Options_V2.php
+++ b/classes/Options_V2.php
@@ -1213,12 +1213,13 @@ class Options_V2 {
 								'desc'    => __( 'Define how many columns you want to use to display courses.', 'tutor' ),
 							),
 							array(
-								'key'         => 'course_archive_filter',
-								'type'        => 'toggle_switch',
-								'label'       => __( 'Course Filter', 'tutor' ),
-								'label_title' => '',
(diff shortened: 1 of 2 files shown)


classes/Options_V2.php
```diff
@@ -1213,12 +1213,13 @@ public function get_setting_fields() {
...
-								'label_title' => '',
-								'default'     => 'off',
-								'desc'        => __( 'Show sorting and filtering options on course archive page', 'tutor' ),
+								'key'           => 'course_archive_filter',
+								'type'          => 'toggle_switch',
+								'label'         => __( 'Course Filter', 'tutor' ),
+								'label_title'   => '',
+								'default'       => 'off',
+								'desc'          => __( 'Show sorting and filtering options on course archive page', 'tutor' ),
+								'toggle_fields' => 'courses_per_page,supported_course_filters,course_archive_filter_sorting',
```
sazedul-haque: This should toggle only `supported_course_filters`. No need to change any JavaScript. 
Also, move the `courses_per_page` block before `course_archive_filter`.

PR #2059, opened 2025-08-21
⚡ Improve quiz attempt list queries and result processing performance

harunollyo:
**Observation**
The current quiz attempt list uses a complex and inefficient approach: it processes attempts at runtime, counting each result individually. This leads to slower performance and adds unnecessary system overhead.

**Solution**
- Removed redundant/unnecessary code
- Simplified and optimized database queries for better performance
- Implemented a background cron job to migrate old attempt results behind the scenes, ensuring no impact on the user experience


migrations/QuizAttemptMigrator.php
```diff
@@ -0,0 +1,99 @@
...
+	protected $schedule_interval = 5;
+
+	/**
+	 * Get total unprocessed result.
+	 *
+	 * @return int
+	 */
+	protected function get_total_items() {
+		global $wpdb;
+		return (int) $wpdb->get_var( "SELECT COUNT(*) FROM {$wpdb->prefix}tutor_quiz_attempts WHERE result IS NULL" );
```
shewa12: We can use the query helper!

classes/Quiz_Attempts_List.php
```diff
@@ -157,80 +121,46 @@ public function get_quiz_attempts_stat() {
...
+				$search_term_raw = $search;
+				$search_filter   = '%' . $wpdb->esc_like( $search ) . '%';
 
-				$count_obj->pending = (int) $wpdb->get_var(
+				//phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+				$results = $wpdb->get_results(
 					$wpdb->prepare(
-						"{$select_stmt}		
-						WHERE attempt_status != %s
+						"SELECT result, COUNT( DISTINCT attempt_id) AS total
```
shewa12, line 135: We can use the query helper!

migrations/BatchProcessor.php
```diff
@@ -0,0 +1,257 @@
+<?php
```
shewa12, line 1: The since tag is missing on all the member variables & methods.

migrations/BatchProcessor.php
```diff
@@ -0,0 +1,257 @@
...
+	/**
+	 * Get items to process.
+	 * Must be implemented in child class.
+	 *
+	 * @param int $offset offset.
+	 * @param int $limit limit.
+	 *
+	 * @return array
+	 */
+	abstract protected function get_items( $offset, $limit);
```
shewa12: All the **Abstract** methods should have argument & return strict type so that the concrete classes implement it properly and the `process_batch` method can work as expected.

migrations/Migration.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Class Migration
+ */
+class Migration {
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		QuizAttemptMigrator::instance()->schedule();
```
shewa12: Once the migration is complete, you no longer need to initiate this process. Need to handle that scenario.

migrations/QuizAttemptMigrator.php
```diff
@@ -0,0 +1,99 @@
...
+	/**
+	 * Get items to batch process.
+	 *
+	 * @param int $offset offset.
+	 * @param int $limit limit.
+	 *
+	 * @return array
+	 */
+	protected function get_items( $offset, $limit ) {
+		global $wpdb;
```
shewa12, line 80: We can use the query helper!

shewa12 (changes requested):
Please check the request changes.

migrations/Migration.php
harunollyo: This is already handled inside BatchProcessor class

migrations/QuizAttemptMigrator.php
harunollyo: Can't. Query helper `get_all` method has no support for offset and this method already used in tons of places.

classes/Quiz_Attempts_List.php
harunollyo: Unfortunately, we can't use the query helper here because it doesn't support `GROUP BY result`

shewa12 (approved):

PR #2065, opened 2025-08-25
Add custom option styling and label content to FormSelectInput

sazedul-haque:
Introduces an optional 'optionItemCss' prop to FormSelectInput for custom styling of select options and adds support for 'labelContent' in the Option type, allowing more flexible rendering of option labels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,10 +41,9 @@ vendor/
 yarn.lock
 .env
 # Ignore Cypress screenshots
-cypress/screenshots/
+cypress/screenshots/*
 # Ignore Cypress fixtures
-cypress/fixtures/course.json
-cypress/fixtures/assets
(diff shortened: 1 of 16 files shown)


assets/react/v3/shared/utils/types.ts
```diff
@@ -42,6 +42,7 @@ export function isObject<T>(value: T): value is T {
 
 export interface Option<T> {
   label: string;
+  labelContent?: ReactNode | string;
   value: T;
```
b-l-i-n-d, line 46: Instead of adding the `optionItemCss` in `FormSelectInput`, can we add the `css` props here? This way, we have more granular control.

sazedul-haque: We don't need different style for every option.

PR #2070, opened 2025-08-26
Models Added To Handle Order Item Meta, Cart Meta, Filter Hook & Refactors

shewa12:


classes/Icon.php
```diff
@@ -16,230 +16,231 @@
...
-	const WARNING                    = 'warning';
-	const WEIGHT_BOX                 = 'weight-box';
-	const XLS                        = 'xls';
-	const XML                        = 'xml';
-	const YOUTUBE                    = 'youtube';
-	const ZIP                        = 'zip';
-	const ZOOM                       = 'zoom';
-	const ZOOM_COLORIZE              = 'zoom-colorize';
+    const ACTIVE = 'active';
+    const ADDONS = 'addons';
```
harunollyo, line 20: CS issue

classes/Icon.php
```diff
@@ -16,230 +16,231 @@
...
+    const ADD_IMAGE = 'add-image';
```
harunollyo, line 21: Same here and more on this Icon.php file

assets/react/helper/utils.js
```diff
@@ -1,3 +1,5 @@
+const { __ } = wp.i18n;
```
sazedul-haque: Need to add this inside the function.

assets/react/helper/utils.js
```diff
@@ -12,3 +14,25 @@ export async function fetchCountriesData() {
...
+ * @returns {string|null} - The Tutor-Message header value or null if not set
+ */
+export async function getTutorMessage(response) {
+    const defaultErrorMessage = __('Something went wrong, please try again', 'tutor-pro');
+    
+    if (!(response instanceof Response)) {
+        return defaultErrorMessage;
+    }
+
+    const res = await response.json();
```
sazedul-haque: We can pass the `res` to `getTutorMessage`. No need to pass the Promise.

sazedul-haque (changes requested):

models/CartModel.php
```diff
@@ -38,12 +38,16 @@ public function __construct() {
 	/**
 	 * Add a course to the user's cart.
 	 *
-	 * @param int $user_id User ID.
-	 * @param int $course_id Course ID.
+	 * @since 3.8.0 $items_details Param added
```
harunollyo: It should be param `$item_type` and `$item_details` added

models/OrderItemMetaModel.php
```diff
@@ -0,0 +1,156 @@
...
+	 * @param string|null $meta_key Meta key (optional).
+	 * @param bool        $single Get a single value or all.
+	 *
+	 * @return array|false Meta results or false if none.
+	 */
+	public function get_meta( int $item_id, $meta_key = null, $single = true ) {
+		$where = array(
+			'item_id' => $item_id,
+		);
+		if ( $meta_key ) {
```
harunollyo, line 80: unnecessary double if check on `meta_key`

models/OrderItemMetaModel.php
```diff
@@ -0,0 +1,156 @@
...
+			if ( $meta_key ) {
+				$where['meta_key'] = sanitize_key( $meta_key );
```
harunollyo: This may cause inconsistent meta_key values if keys are defined with an uppercase or mixed-case suffix like post_ID or anything ending in _ID.

Such keys will be automatically converted to lowercase (e.g., post_ID → post_id).

models/OrderModel.php
```diff
@@ -635,7 +644,7 @@ public function get_order_items_by_id( $order_id ) {
 
 		$where = array( 'order_id' => $order_id );
 
-		$select_columns = array( 'oi.item_id AS id', 'oi.regular_price', 'oi.sale_price', 'oi.discount_price', 'oi.coupon_code', 'p.post_title AS title', 'p.post_type AS type' );
+		$select_columns = array( 'oi.id AS primary_id', 'oi.item_id AS id', 'oi.regular_price', 'oi.sale_price', 'oi.discount_price', 'oi.coupon_code', 'p.post_title AS title', 'p.post_type AS type' );
```
harunollyo, line 647: The current changes may introduce confusion.
It would be clearer and more consistent if we keep this behavior as it was before.

shewa12: If it worked without changes, then would I have made a change there? I need the id & since the item_id as treated as id, I had to give it another name.

PR #2076, opened 2025-08-28
Refactor iframe transform styles for device switcher

nur-alam:


assets/react/admin-dashboard/template-import-script.js
```diff
@@ -56,10 +56,8 @@ document.addEventListener('DOMContentLoaded', function () {
 				iframe.style.width = width;
 				if ('desktop' !== device) {
 					iframe.style.transform = 'scale(0.8085714286)';
-					iframe.style.transformOrigin = 'center top';
```
sazedul-haque, line 59: This will always keep the iframe on left position.

sazedul-haque (changes requested):

PR #2078, opened 2025-08-29
Remove DRY violation and optimize query performance

harunollyo:


helpers/QueryHelper.php
```diff
@@ -1005,40 +1010,20 @@ public static function get_joined_data(
 		$from_clause     = self::prepare_table_name( $primary_table );
 		$join_clauses    = self::build_join_clause( $joining_tables );
 		$where_clause    = self::build_where_search_clause( $where, $search );
-		$order_by_clause = ! empty( $order_by ) ? "ORDER BY {$order_by} {$order}" : '';
+		$order_by_clause = self::build_order_clause( $order_by, $order );
```
shewa12: Some method name starts with build, and some with prepare. Make it consistent with either build or prepare.

helpers/QueryHelper.php
```diff
@@ -620,8 +620,13 @@ protected static function build_order_clause( $orderby = '', $order = 'DESC' ) {
 			return '';
 		}
 
+		// Allow only safe patterns: letters, numbers, dot, underscore.
+		if ( ! preg_match( '/^[a-zA-Z0-9._]+$/', $orderby ) ) {
```
shewa12: It seems the regex is allowing only numbers like: 123 in order by which should not be allowed. Allowed patterns are: letters, letters_with_num, dot or underscore.

helpers/QueryHelper.php
```diff
@@ -1005,40 +1010,20 @@ public static function get_joined_data(
 		$from_clause     = self::prepare_table_name( $primary_table );
 		$join_clauses    = self::build_join_clause( $joining_tables );
 		$where_clause    = self::build_where_search_clause( $where, $search );
-		$order_by_clause = ! empty( $order_by ) ? "ORDER BY {$order_by} {$order}" : '';
+		$order_by_clause = self::build_order_clause( $order_by, $order );
+		$limit_clause    = ( empty( $limit ) && empty( $offset ) ) ? '' : $wpdb->prepare( 'LIMIT %d OFFSET %d', $limit, $offset );
```
shewa12, line 1014: Why do we need to use empty here? !$limit is similar to empty($limit)

shewa12 (approved):

PR #2090, opened 2025-09-04
🆕 : New WebhookUrl For all gateway

sanjana4khan:


ecommerce/PaymentHandler.php
```diff
@@ -38,20 +38,10 @@ public function __construct() {
...
-				'callback'            => array( $this, 'handle_ecommerce_webhook' ),
-				'permission_callback' => '__return_true', // Allows public access to the route.
-			)
-		);
 
-		// Register route for authorizenet.
 		register_rest_route(
 			'tutor/v1',
-			'/ecommerce-webhook/authorizenet',
+			'/ecommerce-webhook(/(?<payment_gateway>[a-z0-9_-]+))?',
```
shewa12: Make this name `payment_method` as it previously.

ecommerce/PaymentHandler.php
```diff
@@ -80,14 +70,16 @@ public function handle_ecommerce_webhook( WP_REST_Request $request ) {
 		);
 		// phpcs:enable
 
-		$route          = $request->get_route() ?? null;
-		$payment_method = ! is_null( $route ) && strpos( $route, self::AUTHORIZENET ) ? self::AUTHORIZENET : Input::get( 'payment_method', 'paypal' );
+		$payment_method = $request->get_param( 'payment_method' ) ?? $request->get_param( 'payment_gateway' ) ?? '';
```
shewa12: After changing the above we can use just this: `$request->get_param( 'payment_method' )`

ecommerce/PaymentHandler.php
sanjana4khan: Done

ecommerce/PaymentHandler.php
sanjana4khan: done

PR #2100, opened 2025-09-12
✨ Certificate added to course bundles

b-l-i-n-d:
Moved Certificate import to shared components and updated its usage to pass currentCertificateKey, onSelect handler, and certificateTemplates props. This enables certificate selection and template management in the Additional page.


classes/Utils.php
shewa12: Need to refactor this method.

PR #2101, opened 2025-09-12
Refactor: restructure profile data array for improved readability and maintainability

harunollyo:
- A new hook also introduced to extend functionality

### DIFF
--- a/templates/dashboard/my-profile.php
+++ b/templates/dashboard/my-profile.php
@@ -59,13 +59,13 @@ $profile_data = array(
 	),
 );
 
+$profile_data = apply_filters( 'tutor_profile_data', $profile_data );
 ?>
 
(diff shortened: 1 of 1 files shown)


templates/dashboard/my-profile.php
```diff
@@ -25,33 +25,60 @@
...
+	'skill_occupation'  => array(
+		'label' => __( 'Skill/Occupation', 'tutor' ),
+		'value' => $job ? $job : '-',
+	),
+	'biography'         => array(
+		'label' => __( 'Biography', 'tutor' ),
+		'value' => $bio ? $bio : '-',
+	),
 );
+
```
shewa12, line 61: We should provide a filter hook to the `$profile_data` variable. It will allow others to add custom information, instead of providing a hook to modify HTML.

shewa12 (approved):

PR #2113, opened 2025-09-19
Fix: show total marks based on final grade calculation settings

Anindra123:


views/quiz/attempt-table.php
```diff
@@ -21,7 +21,7 @@
 if ( 'course-single-previous-attempts' == $context && is_array( $attempt_list ) && count( $attempt_list ) ) {
 	// Provide the attempt data from the first attempt.
 	// For now now attempt specific data is shown, that's why no problem if we take meta data from any attempt.
```
harunollyo: Update the comment and describe what exactly doing here

PR #2121, opened 2025-09-23
Fix(🛠️) : Pay option for manual enrollment.

sanjana4khan:


ecommerce/CheckoutController.php
```diff
@@ -1128,4 +1135,29 @@ protected function validate_pay_now_req( array $data ) {
...
+	 * @return array{
+	 *     courses: array{
+	 *         total_count: int,
+	 *         results: \WP_Post[]
+	 *     }
+	 * }
+	 */
+	public function get_courses_data_by_order_id( $order_items ): array {
+
+		$results = array_map( fn( $item ): \WP_Post => get_post( $item->id ), $order_items );
```
shewa12: Since the arrow function has a returned type declared, it will throw an error if get_post returns a **null** value. This case needs to be handled.

shewa12 (changes requested):

PR #2147, opened 2025-10-16
Fix: arbitrary assignment submit list can access

harunollyo:


templates/dashboard/assignments/submitted.php
```diff
@@ -18,8 +18,15 @@
 use TUTOR_ASSIGNMENTS\Assignments_List;
 
 $order_filter          = Input::get( 'order', 'desc' );
-$assignment_id         = Input::get( 'assignment' );
+$assignment_id         = Input::get( 'assignment', 0, Input::TYPE_INT );
 $format                = get_option( 'date_format' ) . ' ' . get_option( 'time_format' );
+$course_id             = tutor_utils()->get_course_id_by( 'assignment', $assignment_id );
+
+if ( ! tutor_utils()->can_user_edit_course( get_current_user_id(), $course_id ) ) {
+	tutor_utils()->tutor_empty_state();
```
shewa12, line 26: If the user doesn't have permission, then it can show unauthorized access instead of an empty state.

PR #2152, opened 2025-10-21
Fix(🛠️) : Paypal security issue

sanjana4khan:


ecommerce/PaymentGateways/Paypal/src/Payments/Paypal/Api.php
```diff
@@ -0,0 +1,281 @@
+<?php
```
shewa12, line 1: Add the doc comment for this class

ecommerce/PaymentGateways/Paypal/src/Payments/Paypal/Helper.php
```diff
@@ -0,0 +1,421 @@
...
+			}
+
+			if ( ! empty( $errorBody->message ) ) {
+				$message .= "{$errorBody->name} : {$errorBody->message}. ";
+			}
+
+			return $message;
+		}
+
+		return null;
```
shewa12, line 207: Since the return type is declared string on the doc comment it should return string instead of `null`

shewa12 (approved):

PR #2187, opened 2025-11-04
V4 pagination

nur-alam:


assets/core/scss/components/_pagination.scss
```diff
@@ -0,0 +1,33 @@
+// Pagination scss
+
+@use '../tokens' as *;
+@use '../mixins' as *;
+
+.tutor-pagination {
+	@include tutor-pagination-base;
+}
+
+.tutor-pagination__info {
```
sazedul-haque: Please use single hyphen (-) instead of double underscore (__). And for styles for child elements, nested style is more preferable.

templates/demo-components/nur.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
sazedul-haque: Why nur components? Let's keep everything inside dashboard.

templates/demo-components/dashboard.php
```diff
@@ -1 +1,51 @@
...
+ *
+ * @package Tutor
+ * @since 1.0.0
+ */
+
+use TUTOR\Icon;
+
+?>
+
+<h2>Dashboard</h2>
```
sazedul-haque: This should not be inside dashboard.

sazedul-haque (changes requested):

sazedul-haque (approved):

PR #2255, opened 2025-12-02
Security Issue Fixes for 3.9.4

Anindra123:
# Security Issue Fix for 3.9.4

## Issue 1
 - Inside `CouponController` class for both ajax method `bulk_action_handler` and `coupon_permanent_delete` there is  a check for user capability however the error message is not returned, the method `tutor_utils()->error_message()` is called but not returned thus it is not returning from the methods
 
 - To fix this i have use the method `tutor_utils()->check_current_user_capability()` which correctly checks user capability and returns a json error if unauthorized.
 
 ## Issue 2
 - On the `pay_now` action which is called during checkout you can inject SQL on the `coupon_code` argument when passing the form data during checkout
 - This in-turn runs a SQL query when trying to query the data of the coupon code by calling the `get_coupon` method inside `CouponModel` class
 - This calls the `QueryHelper::get_row` method which takes a where clause, and when preparing the where clause the malicious SQL code gets added.
 
 - To fix this i have put coupon_code inside an `esc_sql` inside `get_coupon_details_for_checkout` method which calls the `get_coupon` method and passes the coupon_code
 
 ## Issue 3
 - On the `tutor_order_details` action the method it calls has nonce check but no capability check
 - After looking over everything this action was called i have found out that it is only used for admin users not any other users
 
 - So to fix it i added `tutor_utils()->check_current_user_capability()` method so that only admin can view the order details
 
 ## Issue 4
 - On `mark_course_complete()` method no check was provided whether user is enrolled, allowing any user to mark course as complete even if it is paid course
 
 - To fix it added a check if user is enrolled using `tutor_utils()->is_enrolled()`, if not then exit from the function
 
 
 ## Issue 5
 - On `course_enrollment()` method there is no check if the course is purchasable allowing user to enroll in course with ajax call.
 
 - To fix this i added the `tutor_utils()->is_course_purchasable()` check to check if user is need to purchase the course before enrollment
 
 ## Issue 6
 - On reviewing quiz attempt details by instructor on the backend admin panel, any attempt done by students in a course of one instructor can be viewed by other instructor simply by copying the link
 - To fix this i obtained the `course_id` from the attempt details and check if the course id belong to the specific instructor using the method `tutor_utils()->is_instructor_of_this_course`
 
 - On the `tutor_instructor_feedback` ajax call any other instructor can add their own feedback
 - To fix this i added the same check by getting the `course_id` from attempt details and checking if the attempt is in a course by the instructor

### DIFF
--- a/classes/Course.php
+++ b/classes/Course.php
@@ -2999,14 +2999,12 @@ class Course extends Tutor_Base {
 			}
 
 			/**
-			 * This check was added for a security issue
-			 * where users without purchasing a course can
-			 * enroll into the course by ajax call. To fix this
-			 * the following check was added to check if course is
-			 * paid or not. Furthermore, to check if user has already
-			 * purchased the course another check was added to check
-			 * if user is enrolled in course cause default behavior
-			 * for tutor is user is enrolled in course if purchased.
+			 * This check was added to address a security issue where users could
+			 * enroll in a course via an AJAX call without purchasing it.
+			 *
+			 * To prevent this, we now verify whether the course is paid.
+			 * Additionally, we check if the user is already enrolled, since
+			 * Tutor's default behavior enrolls users automatically upon purchase.
 			 *
 			 * @since 3.9.4
 			 */
--- a/classes/Course.php
+++ b/classes/Course.php
@@ -2806,10 +2806,10 @@ class Course extends Tutor_Base {
 					}
 				}
 				if ( ! $has_passed ) {
-					++$required_assignment_pass;
+					$required_assignment_pass++;
 				}
 			} else {
-				++$required_assignment_pass;
+				$required_assignment_pass++;
 			}
 		}
 
--- a/classes/Course.php
+++ b/classes/Course.php
@@ -2992,16 +2992,35 @@ class Course extends Tutor_Base {
 		$course_id = Input::post( 'course_id', 0, Input::TYPE_INT );
 		$user_id   = get_current_user_id();
 
-		$is_enrolled = (bool) tutor_utils()->is_enrolled( $course_id, $user_id );
-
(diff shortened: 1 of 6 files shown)


ecommerce/OrderController.php
```diff
@@ -259,9 +259,8 @@ public function create_order( int $user_id, array $items, string $payment_status
 	 * @return void
 	 */
 	public function get_order_by_id() {
```
harunollyo: This method is bind with ajax `add_action( 'wp_ajax_tutor_order_details', array( $this, 'get_order_by_id' ) );` and confusing with model method name. Rename it like `ajax_get_order_details` to enhance clarity of it's action.

classes/Course.php
```diff
@@ -2125,6 +2125,11 @@ public function mark_course_complete() {
 		 * from the filter value to prevent the completion.
 		 */
 		$can_complete = apply_filters( 'tutor_user_can_complete_course', true, $user_id, $course_id );
+
+		if ( ! tutor_utils()->is_enrolled( $course_id, $user_id ) ) {
```
harunollyo: Move this check before filter hook.

classes/Course.php
```diff
@@ -2992,6 +2997,11 @@ public function course_enrollment() {
 			if ( $password_protected ) {
 				wp_send_json_error( __( 'This course is password protected', 'tutor' ) );
 			}
+
+			if ( tutor_utils()->is_course_purchasable( $course_id ) ) {
```
harunollyo, line 3011: This will always fail for paid course

harunollyo (changes requested):

harunollyo (approved):

PR #2299, opened 2026-01-06
Order list functionality implemented under billing section

harunollyo:
- Billing list with sorting
- Attr implemented in Pagination class
- Existing helpers updated to meet current requirement

### DIFF
--- a/templates/dashboard/account/billing/order-history-card.php
+++ b/templates/dashboard/account/billing/order-history-card.php
@@ -12,41 +12,16 @@
 defined( 'ABSPATH' ) || exit;
 
 use Tutor\Ecommerce\Ecommerce;
+use Tutor\Helpers\ComponentHelper;
 use Tutor\Helpers\DateTimeHelper;
 use TUTOR\Icon;
 use Tutor\Models\OrderModel;
-
-$render_status_badge = function ( $status ) {
-	$badge_class = 'secondary';
-
-	if ( 'completed' === $status ) {
-		return '';
-	}
-
-	switch ( $status ) {
-		case 'processing':
-		case 'pending':
-		case 'on-hold':
-			$badge_class = 'pending';
-			break;
-		case 'refunded':
-		case 'cancelled':
-			$badge_class = 'cancelled';
-			break;
-		case 'incomplete':
-			$badge_class = 'secondary';
-			break;
-	}
-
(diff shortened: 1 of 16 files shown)


templates/dashboard/account/billing/order-history-card.php
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+use Tutor\Ecommerce\Ecommerce;
+use Tutor\Helpers\DateTimeHelper;
+use TUTOR\Icon;
+use Tutor\Models\OrderModel;
+
+$render_status_badge = function ( $status ) {
```
shewa12: We can create a reusable function by putting this fn inside a class. Besides, we can use class constants instead of writing hard-coded.

templates/dashboard/account/billing/order-history-card.php
```diff
@@ -0,0 +1,105 @@
...
+		case 'refunded':
+		case 'cancelled':
+			$badge_class = 'cancelled';
+			break;
+		case 'incomplete':
+			$badge_class = 'secondary';
+			break;
+	}
+
+	return '<span class="tutor-capitalize tutor-badge tutor-badge-small tutor-badge-circle tutor-py-none tutor-badge-' . $badge_class . '">' . esc_html( $status ) . '</span>';
```
shewa12: We should use the **Badge** component.

templates/dashboard/account/billing/order-history.php
```diff
@@ -0,0 +1,50 @@
...
+</div>
+
+<div class="tutor-flex tutor-flex-column tutor-gap-4 tutor-order-history">
+	<?php foreach ( $orders as $order ) : //phpcs:ignore ?>
+		<?php include tutor_get_template( 'dashboard.account.billing.order-history-card' ); ?>	
+	<?php endforeach; ?>
+</div>
+
+<?php Pagination::make()
+	->attrs( array( 'class' => 'tutor-px-6 tutor-py-6 tutor-border-t' ) )
```
shewa12: For a single attr we can use attr instead of attrs.

shewa12 (changes requested):
Please see the request changes.

shewa12 (approved):

PR #2300, opened 2026-01-06
New design implemented for dashboard assignments

sazedul-haque:


classes/Utils.php
```diff
@@ -6922,17 +6922,28 @@ public function get_assignments_by_instructor( $instructor_id = 0, $filter_data
 
 		$pagination_query = '';
 		$date_query       = '';
+		$search_query     = '';
 		$sort_query       = 'ORDER BY ID DESC';
 
 		if ( $this->count( $filter_data ) ) {
 			extract( $filter_data );//phpcs:ignore
 
+			if ( ! empty( $search_filter ) ) {
```
harunollyo, line 6933: `$search_filter` is not passed by `$filter_data

classes/Utils.php
```diff
@@ -6922,17 +6922,28 @@ public function get_assignments_by_instructor( $instructor_id = 0, $filter_data
...
+			}
+
 			if ( ! empty( $course_id ) ) {
 				$in_course_ids = $course_id;
 			}
-			if ( ! empty( $date_filter ) ) {
-				$date_filter = tutor_get_formated_date( 'Y-m-d', $date_filter );
-				$date_query  = " AND DATE(post_date) = '{$date_filter}'";
+			if ( ! empty( $start_date ) && ! empty( $end_date ) ) {
+				$start_date = tutor_get_formated_date( 'Y-m-d', $start_date );
```
harunollyo, line 6945: Same here

components/Avatar.php
```diff
@@ -79,6 +86,22 @@ class Avatar extends BaseComponent {
 	 */
 	protected $initials = null;
 
+	/**
+	 * User object or ID.
+	 *
+	 * @since 4.0.0
```
harunollyo, line 92: Keep new line after @since tag

components/Avatar.php
```diff
@@ -79,6 +86,22 @@ class Avatar extends BaseComponent {
...
+	 * @var int|object|null
+	 */
+	protected $user = null;
+
+	/**
+	 * Avatar alt text.
+	 *
+	 * @since 4.0.0
```
harunollyo, line 101: same here

components/Avatar.php
```diff
@@ -165,6 +188,90 @@ public function initials( string $initials ): self {
...
+	 * @param int|object $user User ID or object.
+	 * @return $this
+	 */
+	public function user( $user ): self {
+		if ( ! $user ) {
+			return $this;
+		}
+
+		$user_id    = is_object( $user ) ? $user->ID : (int) $user;
+		$cache_key  = 'tutor_avatar_component_user_data_' . $user_id;
```
harunollyo: No need to use `tutor` prefix. It's not mess with other.

Use
```php
$cache_key  = "avatar_component_user_data_{$user_id}";
```

components/Avatar.php
```diff
@@ -165,6 +188,90 @@ public function initials( string $initials ): self {
...
+				'type'     => $avatar_src ? 'image' : 'initials',
+				'initials' => $initials,
+				'alt'      => $name,
+			);
+
+			// Store in cache.
+			TutorCache::set( $cache_key, $data );
+
+			// Apply to current instance.
+			if ( $data['src'] ) {
```
harunollyo: It's always true, because $data has src key

components/Badge.php
```diff
@@ -173,6 +173,27 @@ public function attr( $key, $value ) {
...
+	/**
+	 * Set additional CSS classes.
+	 *
+	 * @since 4.0.0
+	 *
+	 * @param string|array $classes CSS classes.
+	 *
+	 * @return $this
+	 */
+	public function classes( $classes ) {
```
harunollyo: No need this method, We can pass class by `attr` or `attrs`

harunollyo (changes requested):

assets/core/scss/utilities/_transition.scss
```diff
@@ -0,0 +1,90 @@
...
+  0: 0ms,
+  75: 75ms,
+  100: 100ms,
+  150: 150ms,
+  200: 200ms,
+  300: 300ms,
+  500: 500ms,
+  700: 700ms,
+  1000: 1000ms,
+);
```
b-l-i-n-d: Instead of defining the same array twice, we can name the first one as `tutor-transition-values` and use that.

b-l-i-n-d (approved):

classes/Utils.php
sazedul-haque: The `extract( $filter_data );` made these variables available here. No need to define again.

classes/Utils.php
sazedul-haque: Same

PR #2302, opened 2026-01-07
SvgIcon component added for SVG icon rendering

sazedul-haque:


components/SvgIcon.php
```diff
@@ -0,0 +1,216 @@
...
+namespace Tutor\Components;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * SvgIcon Component Class.
+ *
+ * Example usage:
+ * ```
+ * SvgIcon::make()
```
harunollyo, line 24: We have simple helper, which support all
```php
tutor_utils()->render_svg_icon( Icon::CHECK );
```

harunollyo: This additional class will not add value, rather than make increase codebase and inconsistent code use.

harunollyo (changes requested):

shewa12 (approved):

PR #2309, opened 2026-01-13
Feat: resources api integration

saadman30:


classes/Template.php
harunollyo, line 532: Please pull from `4.0.0-dev` to your branch `feat/resources-api-integration` and update the code accordingly.

harunollyo (changes requested):

PR #2322, opened 2026-01-20
Fix(🛠): Trim unicode whitespace (NBSP) from input strings

sanjana4khan:


ecommerce/PaymentGateways/Configs/PaypalConfig.php
```diff
@@ -97,19 +97,19 @@ public function __construct() {
 	}
 
 	public function getClientSecret(): string {
-		return $this->secret_id;
+		return tutor_utils()->remove_white_space( $this->secret_id );
```
harunollyo, line 100: Instead of removing white space from everywhere. Save the data correctly.

harunollyo (changes requested):

PR #2332, opened 2026-01-23
Fix: 404 page appeared after switch view mode

harunollyo:


assets/src/js/frontend/dashboard/header.ts
```diff
@@ -16,7 +16,7 @@ const header = () => {
           onSuccess: (res: AjaxResponse) => {
             window.TutorCore.toast.success(res?.message);
             setTimeout(() => {
-              window.location.reload();
+              window.location.href = window._tutorobject.tutor_frontend_dashboard_url;
```
sazedul-haque: Please use `tutorConfig`.

sazedul-haque (approved):

PR #2350, opened 2026-02-03
User profile

sanjana4khan:


templates/user-profile.php
```diff
@@ -10,21 +10,20 @@
...
-$student_meta     = get_user_meta( $user_id );
+$user             = wp_get_current_user();
+$student_details  = get_userdata( $user->ID );
+$student_meta     = get_user_meta( $user->ID );
 $edit_profile_url = tutor_utils()->tutor_dashboard_url( 'account/settings' );
 ?>
 
 <div class="tutor-profile-card">
 	<div class="tutor-profile-card-header">
-		<?php Avatar::make()->user( $user_id )->size( Size::SIZE_104 )->render(); ?>
```
harunollyo, line 27: We need to use Avatar component here.

harunollyo (changes requested):

PR #2352, opened 2026-02-03
fix(form): Preference

nur-alam:


templates/dashboard/account/settings/preferences.php
```diff
@@ -16,16 +16,17 @@
 use Tutor\Components\Constants\InputType;
 use Tutor\Components\Constants\Size;
 
-// Theme options.
 $theme_options = UserPreference::get_theme_options();
 
 $font_scale_options = UserPreference::get_font_scale_options();
 
 // Load current user preferences to seed the form.
 $user_preferences = ( new UserPreference( false ) )->get_preferences();
```
harunollyo: Make `get_preferences` a static method to eliminate the need to create an instance and a tutor cache support to avoid unnecessary query if it called multiple in same request.

templates/dashboard/account/settings/preferences.php
```diff
@@ -16,16 +16,17 @@
...
 
+$autoload_course_content = (bool) tutor_utils()->get_option( 'autoload_next_course_content' );
+
 $default_values = array(
-	'auto_play_next' => isset( $user_preferences['auto_play_next'] ) ? (bool) $user_preferences['auto_play_next'] : false,
```
harunollyo, line 28: Instead of checking isset on every pref. make a static method with default support
```php
public static function get( $key, $default = false, $user_id = 0 ) {

}

UserPreference::get( 'auto_play_next' );
```

harunollyo (changes requested):

harunollyo (approved):

PR #2353, opened 2026-02-03
Update File Upload Component and Added some endpoints and Icon for Live Classes

Anindra123:


assets/icons/json.svg
b-l-i-n-d, line 1: Use `currentColor`. Else use `colorize` suffix.

components/FileUploader.php
```diff
@@ -277,6 +295,33 @@ public function uploader_icon( $icon ) {
...
 	}
 
+	/**
+	 * Set whether to show subtitle.
+	 *
+	 * @param boolean $show_subtitle show subtitle or not.
+	 *
+	 * @return self
+	 */
+	public function show_subtitle( $show_subtitle = true ) {
```
b-l-i-n-d: If the subtitle field is not needed, just don't pass it. Add support for this feature.

components/Popover.php
```diff
@@ -168,11 +168,13 @@ public function title( string $popover_title ): self {
...
 	 * @param string $popover_body the popover body html.
+	 * @param string $popover_body_esc the escape function.
 	 *
 	 * @return self
 	 */
-	public function body( string $popover_body ): self {
-		$this->popover_body = $popover_body;
+	public function body( string $popover_body, string $popover_body_esc = '' ): self {
+		$this->popover_body     = $popover_body;
+		$this->popover_body_esc = $popover_body_esc;
```
b-l-i-n-d: Use this format

```suggestion
	/**
	 * Sets the confirmation message.
	 *
	 * @since 4.0.0
	 *
	 * @param string $message The confirmation message text.
	 * @param array  $allowed_html_tags Optional. Allowed HTML tags and attributes. Keys are tag names and values are allowed attributes.
	 *
	 * @return $this
	 */
	public function message( string $message, array $allowed_html_tags = array() ) {
		$this->message           = $message;
		$this->allowed_html_tags = wp_parse_args( $allowed_html_tags, $this->allowed_html_tags );
		return $this;
	}
```

b-l-i-n-d (changes requested):

b-l-i-n-d (approved):

components/Popover.php
```diff
@@ -63,7 +63,7 @@ class Popover extends BaseComponent {
 	 *
 	 * @var string
 	 */
-	protected $popover_body_esc = 'wp_kses_post';
+	protected $popover_body_esc = '';
```
harunollyo, line 67: Use default `allowed_html_tags` and pass if extra is required

harunollyo (changes requested):

harunollyo (approved):

PR #2365, opened 2026-02-09
Added prepare statement when building where clause

Anindra123:
### Overview
As a part of security fix in 3.9.4, where sql injection was possible through the `coupon_code` parameter a fix was added where `esc_sql` was called to remove any kind of sql injection, however this was not the correct fix. The correct fix is using prepare statement in `prepare_where_clause` method where any user value can be passed. Therefore, I have updated the method specifically the `make_clause` method that is being used in `prepare_where_clause` only, which creates the final clause, here i have checked whether the value is string or integer and added appropriate prepare statement for each type. I have also added some new unit test cases to check the `prepare_where_clause` output making sure it doesn't break the query. 

### Test Case Results
Screenshot 2026-02-09 at 5 44 05 PM

### DIFF
--- a/helpers/QueryHelper.php
+++ b/helpers/QueryHelper.php
@@ -246,21 +246,6 @@ class QueryHelper {
 	}
 
 	/**
-	 * Prepare BETWEEN and NOT BETWEEN sql clauses.
-	 *
-	 * @since 3.9.7
-	 *
-	 * @param array $values the array of values.
-	 *
-	 * @return string
-	 */
-	public static function prepare_between_clause( array $values ): string {
-		$values = array_map( fn( $val ) => self::prepare_value( $val ), $values );
-		$value  = implode( ' AND ', $values );
-		return $value;
-	}
-
-	/**
 	 * Make tge where clause base on its column operator and values.
 	 *
 	 * If the operator is IN then make the clause like `WHERE column_name IN (value1, value2, ...)`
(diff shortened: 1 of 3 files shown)


helpers/QueryHelper.php
```diff
@@ -252,17 +252,32 @@ public static function insert_multiple_rows( $table, $request, $return_ids = fal
...
 	 *
 	 * @return string
 	 */
 	public static function make_clause( array $where ) {
+		global $wpdb;
 		list ( $field, $operator, $value ) = $where;
 
 		$upper_operator = strtoupper( $operator );
 		if ( in_array( $upper_operator, array( 'IN', 'NOT IN' ), true ) ) {
 			$value = '(' . self::prepare_in_clause( $value ) . ')';
```
harunollyo, line 267: Make a prepare_value and use it
```php
public static function prepare_value($value) {
		global $wpdb;
		$escaped_value = null;
		if ( is_int( $value ) ) {
			$escaped_value = $wpdb->prepare( '%d', $value );
		} else if( is_float( $value ) ) {
			list( $whole, $decimal ) = explode( '.', $value );
			$expression = '%.'. strlen( $decimal ) . 'f';
			$escaped_value = $wpdb->prepare( $expression, $value );
		} else {
			$escaped_value = $wpdb->prepare( '%s', $value );
		}
		return $escaped_value;
	}

/**
	 * Make sanitized SQL IN clause value from an array
	 *
	 * @since 2.1.1
	 * 
	 * @param array $arr a sequential array.
	 * 
	 * @return string
	 */
	public static function prepare_in_clause( array $arr ) {
		$escaped = array_map( fn( $value ) => self::prepare_value( $value ), $arr );
		return implode( ',', $escaped );
	}
```

harunollyo (changes requested):

helpers/QueryHelper.php
```diff
@@ -252,17 +252,32 @@ public static function insert_multiple_rows( $table, $request, $return_ids = fal
...
+		} elseif ( in_array( $upper_operator, array( 'BETWEEN', 'NOT BETWEEN' ), true ) ) {
+			$values = explode( 'AND', $value );
```
harunollyo: We can use it here

harunollyo (changes requested):

harunollyo (approved):

PR #2403, opened 2026-03-02
v4-fixes

Anindra123:
### Instructor Quiz Attempt List Page

- Fixed the issue where after filtering if there is no result the filter also does not show, only the empty state shows
- Added Clear All button to clear the filters
- Style Fixes
- Fixed the issue where after a pagination link is clicked and after that when the date filter is set it does not show any data, this is due to the `current_page` filter is still added on the url. To fix this i check if the date filter is added in the url and if the current count of attempts exceed the max then i set the offset to 0 so that it doesn't give me offset value when querying

### Student Quiz Attempt List Page

- Added Clear All button to clear the filters
- Updated style of dropdown filter `btn-link` class

### DIFF
--- a/assets/src/scss/frontend/components/_quiz-attempts.scss
+++ b/assets/src/scss/frontend/components/_quiz-attempts.scss
@@ -11,10 +11,19 @@
 }
 
 .tutor-quiz-attempts {
-  @include tutor-breakpoint-down(sm) {
-    background-color: $tutor-surface-l1;
-    border: 1px solid $tutor-border-idle;
-    border-radius: $tutor-radius-2xl;
+  .tutor-quiz-students-attempts-filter-item:nth-child(1) {
+    .tutor-btn-link {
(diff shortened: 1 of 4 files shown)


assets/src/scss/frontend/components/_quiz-attempts.scss
```diff
@@ -11,18 +11,27 @@
 }
 
 .tutor-quiz-attempts {
-  @include tutor-breakpoint-down(sm) {
-    background-color: $tutor-surface-l1;
-    border: 1px solid $tutor-border-idle;
-    border-radius: $tutor-radius-2xl;
+  .tutor-quiz-students-attempts-filter-item:nth-child(1) {
```
sazedul-haque: No need this. I have added a fix to the DropdownFilter component.

sazedul-haque (approved):

harunollyo (approved):

PR #2407, opened 2026-03-03
Security Fixes for 3.9.8

Anindra123:
## Issue 1

- The action `tutor_update_course_content_order` which allows to change the order of the lesson inside a topic, a payload can be passed `tutor_topics_lessons_sorting` which takes topic_id and array of lesson id
- No check is given for the payload `tutor_topics_lessons_sorting` on whether the user can edit the topic, thus allowing subscribers to modify topic id, add other lessons etc.
- To fix this i have added the check for the topic id using the method `tutor_utils()->can_user_manage( 'topic', id );`

## Issue 2

- For both `enroll_now` and `course_enrollment` method inside `Course.php` class there is check whether the course is paid course but there is not check whether the course is private, thus users can enroll in private courses through ajax call.

- To fix this i have added private courses check on both of the methods.

### DIFF
--- a/classes/Course.php
+++ b/classes/Course.php
@@ -1703,12 +1703,14 @@ class Course extends Tutor_Base {
 		$sorting_order = Input::post( 'tutor_topics_lessons_sorting', '' );
 		$sorting_order = json_decode( $sorting_order, true ) ?? array();
 
-		if ( tutor_utils()->count( $sorting_order ) ) {
-			foreach ( $sorting_order as $topic ) {
-				if ( isset( $topic['topic_id'] ) && ! tutor_utils()->can_user_manage( 'topic', $topic['topic_id'] ) ) {
-					wp_send_json_error( __( 'Access Denied!', 'tutor' ) );
-					return;
-				}
+		if ( ! tutor_utils()->count( $sorting_order ) ) {
+			wp_send_json_error( __( 'Sorting order is required', 'tutor' ) );
+		}
+
+		foreach ( $sorting_order as $topic ) {
+			if ( isset( $topic['topic_id'] ) && ! tutor_utils()->can_user_manage( 'topic', $topic['topic_id'] ) ) {
+				wp_send_json_error( __( 'Access Denied!', 'tutor' ) );
+				return;
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


classes/Course.php
```diff
@@ -1697,9 +1697,21 @@ public function get_wc_product() {
 	 * @since 1.0.0
 	 * @return void
 	 */
-	public function tutor_update_course_content_order() {
+	public function ajax_update_course_content_order() {
 		tutor_utils()->checking_nonce();
 
+		$sorting_order = Input::post( 'tutor_topics_lessons_sorting', '' );
+		$sorting_order = json_decode( $sorting_order, true ) ?? array();
+
```
harunollyo, line 1705: Use early return
```php
if ( ! tutor_utils()->count( $sorting_order ) ) {
   wp_send_json_error( __( 'Sorting order is required', 'tutor' ) );
}
```

classes/Course.php
```diff
@@ -1763,55 +1775,53 @@ public function restrict_media( $where ) {
...
 	 *
 	 * @since 1.0.0
+	 * @since 3.9.8 param $order added.
+	 *
+	 * @param array $order the lesson order array.
+	 *
 	 * @return void
 	 */
-	private function save_course_content_order() {
+	private function save_course_content_order( $order = array() ) {
```
harunollyo: rename the param `$sorting_order`. `$order` has ambiguity

classes/Course.php
```diff
@@ -1763,55 +1775,53 @@ public function restrict_media( $where ) {
...
 		global $wpdb;
```
harunollyo, line 1788: Use early return
```php
if ( ! tutor_utils()->count( $sorting_order ) ) {
   return;
}
```

classes/Course.php
```diff
@@ -1763,55 +1775,53 @@ public function restrict_media( $where ) {
...
-							$wpdb->update(
-								$wpdb->posts,
-								array(
-									'post_parent' => $topic['topic_id'],
-									'menu_order'  => $lesson_key,
-								),
-								array( 'ID' => $lesson_id )
-							);
-						}
+		if ( tutor_utils()->count( $order ) ) {
```
harunollyo: revert this changes

harunollyo (changes requested):

harunollyo (approved):

PR #2424, opened 2026-03-09
Necessary update for manual subscription payment

harunollyo:


assets/core/ts/services/Location.ts
```diff
@@ -0,0 +1,41 @@
+import Alpine from 'alpinejs';
+import axios from 'axios';
+import { type QueryService, type QueryState, queryServiceMeta } from '@Core/ts/services/Query';
+import { type ServiceMeta } from '@Core/ts/types';
+import { tutorConfig } from '@TutorShared/config/config';
+import endpoints from '@TutorShared/utils/endpoints';
+
+export interface Country {
```
b-l-i-n-d: No need to export.

assets/core/ts/services/Location.ts
```diff
@@ -0,0 +1,41 @@
...
+  name: string;
+  states: { id: number; name: string }[];
+}
+
+export class LocationService {
```
b-l-i-n-d: No need to export.

assets/core/ts/services/Location.ts
```diff
@@ -0,0 +1,41 @@
...
+  private readonly _24_HOURS = 1000 * 60 * 60 * 24;
```
b-l-i-n-d: No need to use the underscore `_` for a read-only variable naming. Also, the name should not depend on the variable's value. Please use a generic name.
```suggestion
  private readonly STEAL_TIME = 1000 * 60 * 60 * 24;
```

b-l-i-n-d (changes requested):

b-l-i-n-d (approved):

PR #2444, opened 2026-03-13
🩹 Fix dashboard account layout

b-l-i-n-d:
This PR fixes the dashboard account/billing layout alignment for profile-related pages.

Previously, these pages used inconsistent container classes and max-width rules, which produced narrower content and misaligned headers compared to other dashboard sections. Users saw mismatched horizontal spacing and layout shifts between account tabs.

Root cause: the account header and billing layout were using a different container class, and some page-specific SCSS still enforced a hard `max-width: 728px`.

Fix: standardize the container class to the profile container and remove the hard max-width rules, letting the shared dashboard layout apply consistently.

Tests: not run (CSS-only changes).

### DIFF
--- a/assets/src/scss/frontend/dashboard/layout/_account.scss
+++ b/assets/src/scss/frontend/dashboard/layout/_account.scss
@@ -16,10 +16,4 @@
       }
     }
   }
-
-  .tutor-account-container {
-    max-width: 728px;
-    margin: 0 auto;
-    padding-inline: $tutor-spacing-6;
-  }
 }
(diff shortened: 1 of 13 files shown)


assets/src/scss/frontend/dashboard/layout/_account.scss
```diff
@@ -16,10 +16,4 @@
       }
     }
   }
-
-  .tutor-account-container {
-    max-width: 728px;
-    margin: 0 auto;
-    padding-inline: $tutor-spacing-6;
-  }
 }
```
b-l-i-n-d:
```suggestion
}

.tutor-account-container {
  @include tutor-frontend-layout();
  padding-inline: $tutor-spacing-6;
}
```

sazedul-haque (approved):

PR #2445, opened 2026-03-13
✨ Add quiz auto-start radar and result modals

b-l-i-n-d:
This PR adds the quiz auto‑start radar countdown and new result modals for timeout and successful submission in the learning area.

Previously, auto‑start happened immediately without user feedback and submissions reloaded the page on success. Users could miss the start moment and had no consistent, modal‑based result summary on completion or timeout.

Root cause: the learning‑area quiz flow lacked a dedicated countdown UI and relied on reload‑based success handling. There was also no centralized modal close signal to reload after the result modal is dismissed.

Fix: introduce a radar countdown Alpine component and an auto‑start modal, replace reload‑on‑success with result modals for submit/timeout, and emit a modal‑closed event to reload when the result modal is dismissed. The result modal uses SVG assets for submitted/timeout states and includes the action button inside the modal body to avoid footer borders.

Tests: not run (manual verification only).
image
image
image


components/Modal.php
```diff
@@ -465,7 +465,7 @@ public function get(): string {
 		$footer = $this->render_footer();
 
 		$this->component_string = sprintf(
-			'<div x-data="tutorModal(%s)" x-cloak>
+			'<div x-data="tutorModal(%s)" x-cloak style="display: none;">
```
sazedul-haque, line 469: `x-cloak` should have `display: none` style.

templates/learning-area/quiz/modals/result.php
```diff
@@ -0,0 +1,64 @@
...
+		<?php endif; ?>
+	</div>
+
+	<div class="tutor-flex tutor-flex-column tutor-gap-2 tutor-py-7">
+		<div class="tutor-h5 tutor-font-semibold">
+			<?php echo esc_html( $modal_title ); ?>
+		</div>
+
+		<div class="tutor-p3 tutor-text-secondary">
+			<?php if ( $show_attempts ) : ?>
```
sazedul-haque, line 39: We should use sprintf here to make the string translatable properly.

sazedul-haque (approved):

components/Modal.php
b-l-i-n-d: It has, but without the `display: none` it takes space if you use `height: fit-content`

PR #2450, opened 2026-03-16
Fix: Added default upload type for assignment

Anindra123:
- User was able to upload any file type, which caused the file to not show after upload, since those file are not being uploaded as it is incompatible with wordpress file type
- To fix this i have added required file type that user can upload, if any wrong file type is uploaded it will show error


classes/Assets.php
```diff
@@ -179,38 +179,45 @@ function ( $file ) {
...
+			'tutor_frontend_dashboard_url'  => tutor_utils()->get_tutor_dashboard_page_permalink(),
+			'wp_date_format'                => tutor_js_date_format_against_wp(),
+			'start_of_week'                 => get_option( 'start_of_week', 1 ),
+			'is_admin'                      => is_admin(),
+			'is_admin_bar_showing'          => is_admin_bar_showing(),
+			'addons_data'                   => tutor_utils()->prepare_free_addons_data(),
+			'current_user'                  => $current_user,
+			'content_change_event'          => 'tutor_content_changed_event',
+			'is_tutor_course_edit'          => isset( $_GET['action'] ) && 'edit' === $_GET['action'] && tutor()->course_post_type === get_post_type( get_the_ID() ) ? true : false,
+			'assignment_max_file_allowed'   => 'tutor_assignments' === $post_type ? (int) tutor_utils()->get_assignment_option( $post_id, 'upload_files_limit' ) : 0,
```
harunollyo: Move keys to assignment addon

harunollyo (changes requested):

harunollyo:
Need to fix this PR first
https://github.com/themeum/tutor-pro/pull/1592

PR #2467, opened 2026-03-30
🐛 Align WP editor styling

b-l-i-n-d:
## Summary
This change aligns the PHP-rendered WordPress editor with the styling and TinyMCE asset configuration already used by the React `WPEditor` implementation. It also applies the shared input border radius to the editor chrome through shared SCSS so the editor matches other Tutor form fields.

## Problem
The PHP `WPEditor` path and the React `WPEditor` path were not rendering from the same baseline. In practice that meant the PHP-rendered editor could miss the same TinyMCE content stylesheet configuration used in React, and its editor chrome did not inherit the same radius treatment as the textarea and other input components. Users would see inconsistent editor presentation depending on which implementation rendered the field.

## Root Cause
The React editor explicitly configured TinyMCE with Tutor-specific `content_css` and skin settings, while the PHP component depended mostly on `wp_editor()` defaults. On top of that, the shared input SCSS had no matching rules for the TinyMCE wrapper inside `.tutor-wp-editor-container`, so the PHP editor chrome kept WordPress defaults instead of following Tutor's input tokens.

## Fix
The PHP component now provides the TinyMCE content CSS used by the React editor and preserves nested `tinymce` defaults when custom editor config is passed, so instance-level overrides do not accidentally wipe out the shared defaults. The shared input SCSS now styles `.tutor-wp-editor-container` so the TinyMCE wrapper, toolbar, and status bar use the same input radius token as the textarea path.

## Validation
I verified the PHP component with `php -l components/WPEditor.php`. The SCSS change was kept in the shared input component layer and should be picked up by the normal asset build for the branch.

### DIFF
--- a/components/WPEditor.php
+++ b/components/WPEditor.php
@@ -13,6 +13,8 @@
 
 namespace Tutor\Components;
 
+use Tutor\Helpers\UrlHelper;
+
 defined( 'ABSPATH' ) || exit;
 
 /**
@@ -274,7 +276,7 @@ class WPEditor extends BaseComponent {
 			'textarea_name' => $this->name,
 			'tinymce'       => array(
 				'skin'        => 'light',
-				'skin_url'    => tutor()->url . 'assets/lib/tinymce/light',
+				'skin_url'    => UrlHelper::asset( 'lib/tinymce/light' ),
 				'content_css' => $this->get_content_css(),
 			),
 		);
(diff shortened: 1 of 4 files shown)


components/WPEditor.php
```diff
@@ -272,6 +272,32 @@ protected function get_default_config() {
 			'quicktags'     => true,
 			'editor_height' => 150,
 			'textarea_name' => $this->name,
+			'tinymce'       => array(
+				'skin'        => 'light',
+				'skin_url'    => tutor()->url . 'assets/lib/tinymce/light',
```
sazedul-haque: We can use `UrlHelper::asset()` here

sazedul-haque (approved):

PR #2487, opened 2026-04-02
Fixed The Slug Not Found Errors

shewa12:


classes/Admin.php
```diff
@@ -76,17 +75,50 @@ public function __construct() {
...
+	 * @since 4.0.0
+	 *
+	 * @param mixed  $upgrader_object Upgrader instance.
+	 * @param array  $options         Extra arguments passed to the hook.
+	 * @param string $basename         Plugin basename.
+	 * @param array  $text_domain         Text domain.
+	 *
+	 * @return void
+	 */
+	public static function set_permalink_reset_flag( $upgrader_object, $options, $basename = '', $text_domain = '' ) {
```
harunollyo: Instead of adding it here, we can move this to the Permalink class and use it in both Tutor and Tutor Pro:
```php
Permalink::set_flag_for_upgrade( $upgrader_object, $options, $basename = '', $text_domain = '' );
```

harunollyo (changes requested):

PR #2488, opened 2026-04-02
Add course reset option and conditional UI

sazedul-haque:
Introduce a new 'course_reset_progress' option (toggle, default off) to Options_V2 to control whether students can reset course progress. Update the learning-area sidebar to read this option, add minor layout classes, and render the reset button and confirmation modal only when the option is enabled, keeping the feature hidden by default.


templates/learning-area/components/sidebar.php
shewa12, line 44: Default value is already false.

PR #2495, opened 2026-04-03
Fix(🛠): Span tag is showing on withdrawal page when WooCommerce is active.

sanjana4khan:


shewa12 (changes requested):
Use kses for showing price instead of updating the method.

PR #2499, opened 2026-04-03
Fix(🛠): Remove HTML markup parameter from tutor_price() for WooCommerce

sanjana4khan:


shewa12 (changes requested):
Use wp_kses instead of wp_kses_post

PR #2502, opened 2026-04-06
🐛 Fix V4 quiz and account review regressions

b-l-i-n-d:
This PR fixes a set of regressions around the V4 quiz flow and related account surfaces.

The main user-facing issues were in three areas. First, V4 quiz questions were not preserving the legacy answer-randomization behavior for all supported question types. That created inconsistent answer ordering between the legacy flow and the new learning area, especially for ordering, matching, and image matching questions. Second, student quiz attempt dates were being rendered with the wrong timezone offset because the stored attempt timestamp was being passed through a GMT-to-user conversion path that did not match the value being stored. Third, the account reviews screen was not loading the correct review dataset for instructors, so instructor-facing review pages could show the wrong results.

The root cause in the quiz flow was that the V4 renderer had a simplified answer preparation path that did not mirror the older per-question-type randomization rules. The fix restores those rules in the shared PHP rendering layer so ordering questions always use randomized answers, matching and image matching keep the fixed side ordered while randomizing the draggable choices, and other question types continue to respect the existing randomization flag. This keeps V4 behavior aligned with the legacy quiz experience.

The date issue came from formatting attempt timestamps with a GMT conversion helper even though the value being formatted in this path was not a GMT timestamp. The fix switches the formatting step to use the existing `DateTimeHelper` creation path and renders the output with the site's configured WordPress date and time formats. This avoids the extra offset while still keeping the output consistent with user settings.

The reviews issue came from the account reviews template always using the student review query path. That meant instructor account review screens were not pulling instructor review data through the correct model path. The fix now branches by current account view, using `get_reviews_by_user()` for student view and `get_reviews_by_instructor()` for instructor view. As a result, the account reviews page now loads the expected reviews for both roles.

Validation was done with targeted PHP syntax checks on the touched files and a review of the branch delta against `4.0.0-dev`. The branch includes the following commits:

- `fix(quiz): preserve question answer randomization in v4`
- `fix(quiz): use configured format for attempt dates`
- `fix(reviews): load instructor account reviews`

### DIFF
--- a/classes/Quiz.php
+++ b/classes/Quiz.php
@@ -1921,6 +1921,9 @@ class Quiz {
 	 */
 	public static function render_question( $question, $index = 0 ) {
 		$question_settings = maybe_unserialize( $question->question_settings );
+		$question_type     = $question->question_type;
+		$rand_choice       = ! empty( $question_settings['randomize_question'] )
+			&& '1' === $question_settings['randomize_question'];
 
 		// Normalize question type + settings.
 		switch ( $question->question_type ) {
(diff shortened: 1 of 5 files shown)


classes/Quiz.php
```diff
@@ -1966,4 +1959,45 @@ public static function render_question( $question, $index = 0 ) {
...
+	 * @return array{question_answers: array, question_randomized_answers: array}
+	 */
+	private static function prepare_question_answers( int $question_id, string $question_type, bool $rand_choice ): array {
+		$question_answers            = QuizModel::get_answers_by_quiz_question( $question_id );
+		$question_randomized_answers = array();
+
+		// Ordering questions always use a randomized answer list.
+		// Matching and image matching keep the drop-zone items ordered and only shuffle the draggable choices.
+
+		if ( 'ordering' === $question_type ) {
```
shewa12, line 1981: Use constants for quiz types instead of hard-coded string.

shewa12 (approved):

PR #2524, opened 2026-04-07
Fix(🛠) : Course Pre-Requisites not working in Modern Mode

sanjana4khan:
**Reproduce steps:** 

- Create two courses A and B.
- Make A the prerequisite course for B.
- Now, after purchasing course B, the user can access its lessons without enrolling in course A.


templates/learning-area/index.php
```diff
@@ -64,6 +64,8 @@
 $course_complete_modal_id = 'tutor-course-complete-modal';
 $course_retake_modal_id   = 'tutor-course-retake-modal';
 
+do_action( 'tutor/course/single/content/before/all', $tutor_course_id, $tutor_current_content_id );
```
harunollyo: Need to use it top

harunollyo (changes requested):

harunollyo (approved):

PR #2531, opened 2026-04-07
🐞 Fixed Multiple Issues

shewa12:
Fixed issues are:
1. Course autocomplete was not working
2. The Quiz page was showing broken when change base permalink from settings


assets/src/js/frontend/learning-area/index.ts
```diff
@@ -18,14 +18,15 @@ const initializeLearningArea = () => {
 
   // Normalize path segments
   const pathSegments = pathname.split('/').filter(Boolean);
+  const { lesson_slug = 'lessons', quiz_slug = 'quizzes', assignment_slug = 'assignments' } = window._tutorobject || {};
```
sazedul-haque: Please use `tutorConfig` instead of  `window._tutorobject || {}`

PR #2534, opened 2026-04-07
Fix: Course creation from admin toolbar not working

Anindra123:
This PR fixes the issue where course creation from the admin toolbar dropdown wasn't working, since tutor.js is not loading in dashboard page, to fix this i have added new common.ts file for instructor dashboard and added the logic for course creation there


assets/src/js/frontend/dashboard/pages/instructor/common.ts
```diff
@@ -0,0 +1,54 @@
...
+  handler.init();
+
+  const createCourseButton = document.querySelector('.tutor-create-new-course');
+  createCourseButton?.addEventListener('click', async (e) => {
+    e.preventDefault();
+    createCourseButton.classList.add('tutor-loading');
+    createCourseButton.setAttribute('disabled', 'true');
+    const target = e.target as HTMLElement;
+    target.innerHTML = 'Creating...';
+    await handler.handleCreateCourse();
```
b-l-i-n-d: Use `try...catch`

b-l-i-n-d (changes requested):

b-l-i-n-d (approved):

PR #2537, opened 2026-04-08
Multiple design issues fixed

sazedul-haque:
- Modal width issue fixed
- Flex center utility added
- Add block option to Button component 
- Remove global margin/padding and list resets
- Use flex center column in EmptyState

### DIFF
--- a/assets/core/scss/main.scss
+++ b/assets/core/scss/main.scss
@@ -25,8 +25,6 @@
 *::before,
 *::after {
   box-sizing: border-box;
-  margin: 0;
-  padding: 0;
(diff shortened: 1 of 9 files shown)


assets/core/scss/main.scss
```diff
@@ -63,34 +61,6 @@ body {
...
-  list-style: none;
-}
-
-li > ol,
-li > ul,
-ol,
-ul {
-  padding: 0;
-}
-
```
b-l-i-n-d, line 93: Please explain why these are being removed?

PR #2544, opened 2026-04-09
Instructor quiz attempt list

harunollyo:


templates/dashboard/quiz-attempts.php
```diff
@@ -114,23 +107,38 @@
...
-			<?php if ( $quiz_attempts_count ) : ?>
 			<div class="tutor-quiz-attempts-list">
 				<?php
-				foreach ( $quiz_attempts_list as $quiz_index => $quiz_attempt ) {
+				foreach ( $quiz_attempts_list as $quiz_attempt ) {
 					$attempts       = $quiz_attempt['attempts'];
 					$attempts_count = count( $attempts );
 
 					tutor_load_template(
 						'dashboard.components.quiz-attempts-group',
```
sazedul-haque, line 135: No need to use `quiz-attempts-group` template for instructor as we are showing quiz items only.

sazedul-haque (changes requested):

templates/dashboard/quiz-attempts.php
harunollyo: It will be change later. User attempts list has an issue
https://app.clickup.com/t/86ex7a3mn

PR #2546, opened 2026-04-09
Fix: Prevented pasting image on assignment submission textarea

Anindra123:
This pr fixes the issue where students where allowed to paste images on textarea for assignment submit.  This was fixed in both legacy and modern. To fix this in legacy in `tutor_front.js` i check if `tinymce` is available, if it is loop through each editor and handle the `onpaste` event and check if user pasted an image. For modern, the `WPEditor` component was updated to have a new property called `pasteImage` which on paste event if it is true the allow pasting image, otherwise it prevents it.

### DIFF
--- a/components/WPEditor.php
+++ b/components/WPEditor.php
@@ -144,6 +144,15 @@ class WPEditor extends BaseComponent {
 	protected $editor_config = array();
 
 	/**
+	 * Whether editor supports pasting image.
+	 *
+	 * @since 4.0.0
+	 *
+	 * @var boolean
+	 */
+	protected $paste_image = true;
+
+	/**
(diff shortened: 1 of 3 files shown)


components/WPEditor.php
```diff
@@ -157,6 +166,20 @@ public function name( $name ) {
...
+	/**
+	 * Set whether to paste images.
+	 *
+	 * @since 4.0.0
+	 *
+	 * @param boolean $should_paste whether to paste images.
+	 *
+	 * @return $this
+	 */
+	public function pasteImage( $should_paste = true ) {
```
b-l-i-n-d: use `snake_case`

PR #2565, opened 2026-04-15
Remove v4 beta notice permanently on close button click

sazedul-haque:


classes/Admin.php
```diff
@@ -136,11 +142,35 @@ public function show_v4_beta_notice() {
...
+	 *
+	 * @since 3.9.9
+	 *
+	 * @return void
+	 */
+	public function dismiss_v4_beta_notice() {
+		if ( ! isset( $_GET['tutor_dismiss_v4_beta_notice'] ) ) {
+			return;
+		}
+
```
harunollyo: Need to check user is admin

harunollyo (changes requested):

classes/Admin.php
sazedul-haque: In that case, other users who has access to the admin panel will see the notice but won't be able to close it.

sazedul-haque: Updated.

shewa12 (approved):

PR #2566, opened 2026-04-15
Quiz attempt review enhancements

b-l-i-n-d:
## Summary
- Add `is_attempt_answer_skipped()` helper method
- Refactor `review_quiz_answer()` to handle skipped questions
- Use `status_badges` array for multiple status badges in question headers
- Add `question_id` data attribute for manual review actions
- Fix status fallback values to use `'incorrect'` consistently
- Show skipped questions in attempt details
- Disable pointer cursor for non-interactive question previews (except scale)

## Commits
- fix(quiz): add is_attempt_answer_skipped() and fix docblock
- feat(quiz): refactor review_quiz_answer to handle skipped questions
- refactor(quiz): use status_badges array for multiple badges
- fix(quiz): correct status fallback values
- feat(quiz): add question_id data attribute for manual review
- fix(quiz): show skipped questions in attempt details
- fix(preview): disable pointer cursor for non-interactive question options


classes/Quiz.php
```diff
@@ -1208,6 +1213,115 @@ private function get_attempt_answer( int $attempt_answer_id ) {
...
+	 * Get attempt answer by attempt and question IDs.
+	 *
+	 * @since 4.0.0
+	 *
+	 * @param int $attempt_id  Attempt ID.
+	 * @param int $question_id Question ID.
+	 *
+	 * @return object|null
+	 */
+	private function get_attempt_answer_by_attempt_and_question( int $attempt_id, int $question_id ) {
```
shewa12, line 1228: There is already a method for fetching attempt: `tutor_utils()->get_attempt( $attempt_id )`

b-l-i-n-d: We still need `get_attempt_answer_by_attempt_and_question()` because creating a placeholder row is only correct when no answer row already exists for the same `attempt_id + question_id`. In the skipped legacy case there is no `attempt_answer_id`, so the review flow first has to look up by `attempt_id` and `question_id`: if a row exists we should update it, and if not we create the missing row and then apply the review. Without that lookup the fallback becomes a blind insert, which can create duplicate attempt-answer rows for questions that already have a saved record.

PR #2594, opened 2026-04-22
🛠️ Resole Preview modal overflow and svg icon size

b-l-i-n-d:


assets/src/js/v3/entries/course-builder/components/modals/QuestionPreviewModal.tsx
```diff
@@ -508,7 +521,7 @@ const styles = {
     flex-direction: column;
     overflow: hidden;
     max-width: ${activeTab === 'mobile' ? '444px' : '1220px'};
-    height: 686px;
+    height: 586px;
```
sazedul-haque: why fix height. we can use `viewport height`.

sazedul-haque (changes requested):

sazedul-haque (approved):

PR #2601, opened 2026-04-23
✨ Polish responsiveness on dashboard

b-l-i-n-d:


assets/src/scss/frontend/dashboard/layout/_profile-header.scss
```diff
@@ -21,6 +21,7 @@
         width: 100%;
         text-align: center;
         text-transform: capitalize;
+        margin-left: 0;
```
sazedul-haque: This should be `margin-inline-start`

sazedul-haque (approved):

PR #2612, opened 2026-04-24
✨ Filter v4 questions if learning mode is legacy

b-l-i-n-d:
Filter out questions if learning mode is legacy.
- Sacle
- Coordinates
- Draw Image
- Pin Image
- Puzzle

### DIFF
--- a/classes/Utils.php
+++ b/classes/Utils.php
@@ -5219,20 +5219,16 @@ class Utils {
 			'quiz_id' => $quiz_id,
 		);
 
+		/**
+		 * Filter to exclude quiz question types
+		 *
+		 * @since 4.0.0
+		 */
 		$exclude_types = apply_filters( 'tutor_filter_unsupported_quiz_question_types', array() );
 
-		/**
-		 * Safely append excluded question types to the database query conditions.
-		 * If a 'NOT IN' condition already exists for question types, it merges the new types
-		 * to prevent overwriting previous exclusions. Otherwise, it creates a new condition.
-		 */
-		if ( count( $exclude_types ) ) {
-			$exclude_types = array_unique( $exclude_types );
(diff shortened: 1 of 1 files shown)


classes/Utils.php
```diff
@@ -5214,17 +5214,33 @@ public function get_quiz_option( $post_id = 0, $option_key = '', $default = fals
...
+		);
+
+		$exclude_types = apply_filters( 'tutor_filter_unsupported_quiz_question_types', array() );
+
+		/**
+		 * Safely append excluded question types to the database query conditions.
+		 * If a 'NOT IN' condition already exists for question types, it merges the new types
+		 * to prevent overwriting previous exclusions. Otherwise, it creates a new condition.
+		 */
+		if ( count( $exclude_types ) ) {
```
shewa12: Use the `tutor_utils()->count( $exclude_types )` method to check if the item is an array before performing count.

classes/Utils.php
```diff
@@ -5214,17 +5214,33 @@ public function get_quiz_option( $post_id = 0, $option_key = '', $default = fals
...
+			$exclude_types = array_unique( $exclude_types );
+			if ( isset( $where['question_type'] ) && is_array( $where['question_type'] ) && 'NOT IN' === $where['question_type'][0] ) {
```
shewa12: `$where['question_type']` check is not required.

classes/Utils.php
```diff
@@ -5583,32 +5599,48 @@ public function get_random_questions_by_quiz( $quiz_id = 0 ) {
...
```
shewa12: The same here.

shewa12 (changes requested):
Please check the comments.

PR #2624, opened 2026-04-28
Fix(🛠) : Upcoming coupons were appearing as expired in the coupon list.

sanjana4khan:
- A new `scheduled` status is added to display upcoming coupons. 
- Also fix an issue where expiry date could be set earlier than the start date in edit mode.


shewa12 (commented):
It would be good to mark the coupon as expired in the database instead of checking during rendering.

PR #2628, opened 2026-04-28
Resolve EDD Enrollment Issues and Refactor Code

harunollyo:


classes/Utils.php
```diff
@@ -1317,75 +1317,6 @@ public function price_type( $course_id = 0 ) {
...
-	 * @since 3.3.0  param $is_complete added to cache key.
-	 * @since 4.0.0  enrollment order_id and product_id added to enrollment info.
-	 *
-	 * @param int  $course_id course id.
-	 * @param int  $user_id user id.
-	 * @param bool $is_complete Whether to enrollment completed or not.
-	 *
-	 * @return array|bool|null|object
-	 */
-	public function is_enrolled( $course_id = 0, $user_id = 0, bool $is_complete = true ) {
```
shewa12, line 1338: Why do we need to relocate the method? Revert all the removed methods from the Utils class. There may have been integrations that use this method. In the documentation, we may also have this method for enrollment.

shewa12 (changes requested):

classes/Utils.php
harunollyo: No need to worry! 
All existing method calls like `tutor_utils()->is_enrolled()` will continue to work, and `EnrollmentModel::is_enrolled()` will also function as expected.

PR #2635, opened 2026-04-29
Order history for EDD monetization

harunollyo:
Included
- Moved method ref updated.
- Missing direct access prevention.
- Used constant instead of hardcoded string


classes/TutorEDD.php
```diff
@@ -44,8 +46,12 @@ public function __construct() {
 		 */
 		add_filter( 'is_course_purchasable', array( $this, 'is_course_purchasable' ), 10, 2 );
 		add_filter( 'get_tutor_course_price', array( $this, 'get_tutor_course_price' ), 10, 2 );
-
 		add_action( 'edd_update_payment_status', array( $this, 'edd_update_payment_status' ), 10, 3 );
+
+		// @since 4.0.0
+		add_filter( 'tutor_order_history_card_template', fn( $template ) => tutor_get_template( 'dashboard.account.billing.edd-order-history-card' ) );
```
shewa12, line 52: Before returning the template, we should check if the file exists.

classes/TutorEDD.php
```diff
@@ -209,4 +215,95 @@ public function edd_update_payment_status( $payment_id, $new_status, $old_status
...
+	public function filter_order_history_status_options( $options, $seleted ) {
+		$url     = get_pagenum_link();
+		$user_id = get_current_user_id();
+
+		$statuses = array_merge( array( 'all' => 'All' ), edd_get_payment_statuses() );
+		$options  = array();
+
+		foreach ( $statuses as $key => $status ) {
+			$params = array(
+				'meta_key' => Course::IS_TUTOR_ORDER_FOR_COURSE_META,
```
shewa12, line 238: If the meta is the flag for identifying the tutor course order, then we can name it tutor_course_order. Current meta naming is confusing & long: IS_TUTOR_ORDER_FOR_COURSE_META

templates/dashboard/account/billing/wc-order-history-card.php
```diff
@@ -31,7 +31,7 @@
 $order_status   = $wc_order->get_status();
 $payment_status = $wc_order->get_status();
 $payment_method = $wc_order->get_payment_method_title();
-$order_date     = $wc_order->get_date_created();
+$order_date     = $wc_order->get_date_created()->date( get_option( 'date_format' ) . ', ' . get_option( 'time_format' ) );
```
shewa12: Is it guaranteed that the get_order_created returns value will have a date prop? If not then we should check before access.

shewa12 (commented):
Please check the comments.

classes/TutorEDD.php
harunollyo: `tutor_get_template` function does that

templates/dashboard/account/billing/wc-order-history-card.php
harunollyo: yes

classes/TutorEDD.php
harunollyo: I did consider it when choosing the name, but I followed the naming convention that had been used previously.

https://github.com/themeum/tutor/pull/2635/changes/BASE..ee64c098f55c603247480598f831bc42c676b130#diff-9595e80f0b9a8389df9703d09d7152c72c5790e10c231cca458d2fe2e163aaacL105

PR #2637, opened 2026-05-04
🛠️ Misc Quiz Fixes

b-l-i-n-d:


components/PreviewTrigger.php
```diff
@@ -91,9 +110,16 @@ public function get(): string {
 			}
 		}
 
+		$url_value = '';
+		if ( true === $this->url ) {
```
sazedul-haque: `elseif` block will never execute here.

sazedul-haque (changes requested):

components/PreviewTrigger.php
b-l-i-n-d: It will execute if the URL is false.

sazedul-haque (approved):

PR #2638, opened 2026-05-04
💅 Update toast styles and animation

b-l-i-n-d:


shewa12 (changes requested):
We shall not change any code now unless we really need it.

PR #2690, opened 2026-05-11
Chore: make preview new quiz interactive

saadman30:


assets/src/js/v3/entries/course-builder/components/modals/QuestionPreviewModal.tsx
```diff
@@ -143,12 +143,35 @@ const QuestionPreviewModal = ({ question, onClose }: QuestionPreviewModalProps)
 
     iframeDocument.body.setAttribute('data-preview-device', activeTab);
 
+    const iframeWin = iframeDocument.defaultView;
+    const tutorPreviewWin = iframeWin as (Window & { _tutorCoordinatesRedrawAll?: () => void }) | null;
+    let coordinatesRedrawTimeout: number | undefined;
+    if (tutorPreviewWin && typeof tutorPreviewWin._tutorCoordinatesRedrawAll === 'function') {
```
b-l-i-n-d: It seems this condition is only bound to the `coordinate` question type. Why not move this logic inside that component?

b-l-i-n-d (changes requested):

b-l-i-n-d (approved):

PR #2701, opened 2026-05-12
Fix design issues

sazedul-haque:
1. Login forms design fixed for both legacy and new
2. Cover photo added to users profile
3. Hide popover when trigger is not visible
4. Prevent duplicate subpage menu items render
5. Plyr control buttons style fixed


assets/src/scss/frontend/dashboard/_profile.scss
```diff
@@ -11,11 +11,19 @@
...
+    background-repeat: no-repeat;
     border-radius: $tutor-radius-lg;
-    margin-bottom: $tutor-spacing-7;
+    margin-bottom: $tutor-spacing-14;
     position: relative;
     height: 184px;
 
+    .tutor-avatar {
+      border: 4px solid $tutor-border-inverse;
+      transform: translateY(92px);
```
b-l-i-n-d: Instead of using a specific value, try using `position: relative;', with top or bottom and translateY. This is safer for different screen sizes.

b-l-i-n-d (approved):

b-l-i-n-d (approved):

PR #2703, opened 2026-05-12
Query optimize

shewa12:


classes/Utils.php
harunollyo: We can use there `$this->count( $instructor )` which exact same of this

classes/Utils.php
```diff
@@ -7161,40 +7159,36 @@ public function is_instructor_of_this_course( $instructor_id = 0, $course_id = 0
...
-				$instructor_id,
-				$course_id
-			)
+			$meta_query_args['relation'] = 'AND';
+			$meta_query                  = new \WP_Meta_Query( $meta_query_args );
+			$instructor_query            = new \WP_User_Query(
+				array(
+					'include'     => array( $instructor_id ),
+					'number'      => 1,
+					'fields'      => 'ID',
```
harunollyo, line 7185: Old query return `umeta_id` but new return user ID? there are 26 places where it's called can be break

harunollyo (changes requested):

PR #2709, opened 2026-05-13
✨ Show current user's review first in list

b-l-i-n-d:
Ensure the current user's review appears at the top of the course reviews. Retrieve the user's review, normalize it to a single object, filter it out of the existing reviews (by comment_ID) to avoid duplicates, reindex the array, and then prepend the user's review so it displays first.


templates/learning-area/subpages/reviews.php
```diff
@@ -38,6 +38,21 @@
 $reviews       = tutor_utils()->get_course_reviews( $tutor_course_id, $offset, $review_per_page, false, array( 'approved' ), $current_user_id );
 $my_rating     = tutor_utils()->get_reviews_by_user( 0, 0, null, false, $tutor_course_id, array( 'approved', 'hold' ) );
 
+if ( ! empty( $my_rating ) ) {
+	$my_review    = is_array( $my_rating ) ? $my_rating[0] : $my_rating;
```
sazedul-haque: confusion name `$my_rating`, `$my_review` for the same thing.

sazedul-haque (changes requested):

harunollyo (approved):

PR #2728, opened 2026-05-18
Fix: User can purchase unpublished courses

Anindra123:
- After user puts a course to cart if instructor puts the course to trash or any other status other than published user can still checkout and purchase the course
- To fix this i have added backend check to make sure only published course are purchasable

### DIFF
--- a/ecommerce/CartController.php
+++ b/ecommerce/CartController.php
@@ -231,18 +231,18 @@ class CartController {
 			);
 		}
 
-		$can_buy = apply_filters( 'tutor_allow_course_enrollment', true, $course_id );
-		if ( ! $can_buy ) {
+		if ( 'publish' !== get_post_status( $course_id ) ) {
 			$this->json_response(
-				__( 'Cannot add to cart, enrollment is currently paused.', 'tutor' ),
+				__( 'Cannot add to cart, course is not published.', 'tutor' ),
 				null,
 				HttpHelper::STATUS_BAD_REQUEST
 			);
 		}
 
-		if ( 'publish' !== get_post_status( $course_id ) ) {
+		$can_buy = apply_filters( 'tutor_allow_course_enrollment', true, $course_id );
+		if ( ! $can_buy ) {
 			$this->json_response(
-				__( 'Cannot add to cart, course is not published.', 'tutor' ),
+				__( 'Cannot add to cart, enrollment is currently paused.', 'tutor' ),
 				null,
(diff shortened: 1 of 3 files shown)


harunollyo:
Discussion required. Need to update previous code also.
@Anindra123

ecommerce/CheckoutController.php
```diff
@@ -669,6 +649,18 @@ public function pay_now() {
...
 					array_push( $errors, sprintf( __( 'Invalid item: %s', 'tutor' ), get_the_title( $object_id ) ) );
 				}
+
+				$status = get_post_status( $object_id );
+				if ( 'publish' !== $status ) {
+					// translators: %s is the course title.
+					array_push( $errors, sprintf( __( '“%s” is no longer available for purchase.', 'tutor' ), get_the_title( $object_id ) ?? '' ) );
+				}
+
+				$can_buy = apply_filters( 'tutor_allow_course_enrollment', true, $object_id );
```
harunollyo: Change the name of filter `tutor_can_purchase_course` with default value `true`.
Return `WP_Error` if the condition is not satisfied.

harunollyo (changes requested):

ecommerce/CartController.php
```diff
@@ -231,8 +231,16 @@ public function add_course_to_cart() {
 			);
 		}
 
-		$can_buy = apply_filters( 'tutor_allow_course_enrollment', true, $course_id );
-		if ( ! $can_buy ) {
+		if ( 'publish' !== get_post_status( $course_id ) ) {
+			$this->json_response(
```
harunollyo:
```php
$this->response_bad_request( __( 'Cannot add to cart, course is not published.', 'tutor' ) );
```

ecommerce/CartController.php
```diff
@@ -231,8 +231,16 @@ public function add_course_to_cart() {
...
+				__( 'Cannot add to cart, course is not published.', 'tutor' ),
+				null,
+				HttpHelper::STATUS_BAD_REQUEST
+			);
+		}
+
+		$can_buy = apply_filters( 'tutor_can_purchase_course', true, $course_id );
+		if ( is_wp_error( $can_buy ) ) {
 			$this->json_response(
```
harunollyo: same here

harunollyo (changes requested):

harunollyo (approved):

PR #2731, opened 2026-05-19
Updated student dashboard time calculation based on course completion percentage

sanjana4khan:
Previously, student dashboard time spent was calculated after a course was fully completed. 
Now, the time is calculated based on the course completion percentage. 
Also, if a course is marked as completed using the `Complete Your Course` button, then the full course duration is counted.

[Associated PR Link](https://github.com/themeum/tutor-pro/pull/1775)


classes/Course.php
```diff
@@ -3455,19 +3455,29 @@ public static function get_total_course_duration( $course_ids ): array {
 		foreach ( $course_ids as $id ) {
 			$duration = tutor_utils()->get_course_duration( (int) $id, true );
 
-			$total_seconds += ( (int) $duration['durationHours'] * 3600 )
-				+ ( (int) $duration['durationMinutes'] * 60 )
+			$completion_percentage = tutor_utils()->is_completed_course( $id )
```
harunollyo: In Flexible Mode, a student can complete the course at any time. For example, if a course contains 10 lessons but the student completes only 2 lessons and then marks the course as completed, the completion percentage should not be considered 100%. In this scenario, the completion percentage should reflect the actual number of lessons completed.

harunollyo (changes requested):

models/CourseModel.php
```diff
@@ -1682,4 +1683,81 @@ private function build_course_progress_item( \WP_Post $post, $user_id, $course_i
...
+	 *
+	 * @param array<int> $course_ids List of course IDs.
+	 *
+	 * @return array{
+	 *     hours: int,
+	 *     minutes: int,
+	 *     seconds: int
+	 * } Total accumulated duration from all given courses.
+	 */
+	public static function get_total_course_duration( $course_ids ): array {
```
harunollyo, line 1701: Add support for single course id and update comment according to code changes
```php
$course_ids = is_array( $course_ids ) ? $course_ids : array( $course_ids );
```

models/CourseModel.php
```diff
@@ -1682,4 +1683,81 @@ private function build_course_progress_item( \WP_Post $post, $user_id, $course_i
...
+		$total_seconds = 0;
+
```
harunollyo: remove unnecessary space from here and other places

harunollyo (changes requested):

harunollyo (approved):

PR #2747, opened 2026-05-22
📑 Content update across platform

b-l-i-n-d:


classes/QuizBuilder.php
```diff
@@ -169,7 +169,7 @@ public function save_question_answers( $question_id, $question_type, $question_a
 			}
 
 			// Save sort order.
-			$answer_order++;
+			++$answer_order;
```
harunollyo: Should not be change

classes/QuizBuilder.php
```diff
@@ -250,16 +250,16 @@ public function save_questions( $quiz_id, $questions ) {
 		$question_order = 0;
 		foreach ( $questions as $question ) {
 			$data_status = isset( $question[ self::TRACKING_KEY ] ) ? $question[ self::TRACKING_KEY ] : self::FLAG_NO_CHANGE;
-			$question_order++;
+			++$question_order;
```
harunollyo: same here

harunollyo (changes requested):

harunollyo (approved):

Issue #2169, opened 2025-10-30
Multi site install error


webangon:
While working on multisite environment , tutor display fatal error. 

`Fatal error: Uncaught Exception: Table 'u136278019_OeK7f.educas_2_educas_users' doesn't exist in /home/u136278019/domains/*****/public_html/sites/educash/wp-content/plugins/tutor/helpers/QueryHelper.php:1120`

PR #2750, opened 2026-05-28
Fixing WordPress Multisite Compatibility Issues in Tutor LMS

commercial-hippie:
This PR resolves multiple database, schema installation, and page-rendering bugs when running Tutor LMS on a WordPress Multisite network (specifically where subsites use a custom table prefix like `wp_2_`).
#### 1. Bypassing Subsite Prefixes for Shared Global Tables
* **Problem**: In a WordPress Multisite, global tables like `wp_users`, `wp_usermeta`, `wp_blogs`, etc., are shared network-wide and do not receive subsite prefixes. Tutor LMS's `QueryHelper::prepare_table_name()` automatically prepends the subsite table prefix (`wp_2_`) to all incoming tables, causing SQL execution failures like `Table 'wp_2_wp_users' doesn't exist`.
* **Fix**: Updated `QueryHelper::prepare_table_name()` to recognize all standard global WordPress tables (e.g. `$wpdb->users`, `$wpdb->usermeta`, `$wpdb->blogs`, etc.) and return them unmodified without prepending the prefix.
#### 2. Resolving Hardcoded Foreign Key Constraint Name Conflicts
* **Problem**: MySQL/InnoDB requires foreign key constraint names to be globally unique within the database schema. Tutor LMS hardcoded constraint names (such as `fk_tutor_itemmeta`, `fk_tutor_ordermeta_order_id`, etc.) in its table creation SQL statements. Activating or upgrading the plugin on a second subsite would fail with `errno: 150 (Foreign key constraint is incorrectly formed)` because the constraint name was already registered by the primary site.
* **Fix**: Modified `Tutor.php` and `Upgrader.php` to dynamically prefix all database constraints with `{$wpdb->prefix}` (e.g., `fk_{$wpdb->prefix}tutor_itemmeta`), preventing schema conflicts across multiple sites.
#### 3. Resolving Instructor Dashboard Render Crash
* **Problem**: The main `/dashboard/` page crashed (showing only the header and failing to render the content/footer) because `WithdrawModel::get_withdraw_summary` hardcoded the query source as `FROM {$wpdb->prefix}users u`. Since there is no site-specific users table (`wp_2_users`), the query threw a fatal SQL exception.
* **Fix**: Corrected the query in `WithdrawModel.php` to reference the global `$wpdb->users` table instead.

### DIFF
--- a/classes/Tutor.php
+++ b/classes/Tutor.php
@@ -936,7 +936,7 @@ final class Tutor extends Singleton {
 			PRIMARY KEY (id),
 			KEY order_id (order_id),
 			KEY meta_key (meta_key),
-			CONSTRAINT fk_tutor_ordermeta_order_id FOREIGN KEY (order_id) REFERENCES {$wpdb->prefix}tutor_orders(id) ON DELETE CASCADE
+			CONSTRAINT fk_{$wpdb->prefix}tutor_ordermeta_order_id FOREIGN KEY (order_id) REFERENCES {$wpdb->prefix}tutor_orders(id) ON DELETE CASCADE
 		) $charset_collate;";
 
 		$order_items_table = "CREATE TABLE {$wpdb->prefix}tutor_order_items (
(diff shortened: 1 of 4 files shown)


commercial-hippie:
Related https://github.com/themeum/tutor/issues/2169

PR #2758, opened 2026-06-02
Minor Changes To Make Multisite Compatible

shewa12:


helpers/QueryHelper.php
```diff
@@ -1268,6 +1268,22 @@ public static function get_table_prefix() {
 	 * @return string
 	 */
 	public static function prepare_table_name( string $table_name ) {
+		global $wpdb;
```
harunollyo: No need to modify this method. In multisite env WP handle the prefix.


```php
// @see method implementation inside class-wpdb.php
public function get_blog_prefix( $blog_id = null )
```

harunollyo (changes requested):

harunollyo:
WordPress Multisite, site/blog ID 1 is the main site, and its tables are stored without a numeric suffix.
For example, if:
```
base_prefix = wp_
```

Blog ID | Posts Table
-- | --
1 | wp_posts
2 | wp_2_posts
3 | wp_3_posts

helpers/QueryHelper.php
```diff
@@ -1253,11 +1254,41 @@ public static function get_last_query(){
 	 *
 	 * @return string
 	 */
-	public static function get_table_prefix() {
+	public static function get_table_prefix( $table_name ) {
 		global $wpdb;
+
+		if ( is_multisite() ) {
```
harunollyo: Use and instead of nested if

helpers/QueryHelper.php
```diff
@@ -1253,11 +1254,41 @@ public static function get_last_query(){
 	 *
```
harunollyo, line 1253: update comment

helpers/QueryHelper.php
```diff
@@ -1253,11 +1254,41 @@ public static function get_last_query(){
...
+		$base_tables = array(
+			'users',
+			'usermeta',
+		);
+
+		if ( in_array( $table_name, $base_tables, true ) ) {
+			return true;
+		}
+
+		$is_base_table = strpos( $table_name, 'wp_users' ) !== false || strpos( $table_name, 'wp_usermeta' ) !== false;
```
harunollyo: Use loop based on `$base_tables`

harunollyo (changes requested):

helpers/QueryHelper.php
```diff
@@ -73,7 +73,8 @@ public static function update( string $table, array $data, array $where ): bool
 		$wpdb->query( $query );
 
 		if ( $wpdb->last_error ) {
-			error_log( $wpdb->last_error );
+			$error_msg = (string) $wpdb->last_error;
```
harunollyo: this prop is string type, why we need to cast it again with another variable to log.
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wpdb.php#L84

harunollyo (changes requested):

PR #2764, opened 2026-06-04
Fix: SQL injection in withdraw history list status filter parameter

Anindra123:


models/WithdrawModel.php
```diff
@@ -111,11 +113,11 @@ public static function get_withdrawals_history( $user_id = 0, $filter = array(),
 		$query_by_status_sql = '';
 		$query_by_user_sql   = '';
 
-		if ( ! empty( $status ) ) {
-			$status = (array) $status;
-			$status = "'" . implode( "','", $status ) . "'";
+		if ( ! empty( $status ) && in_array( $status, array( 'pending', 'approved', 'rejected' ), true ) ) {
```
harunollyo: Use constant, instead of string

harunollyo (changes requested):

harunollyo (approved):

PR #2769, opened 2026-06-05
Fix: Buying or enrolling in permanently deleted course throwing error

Anindra123:


models/CourseModel.php
```diff
@@ -91,6 +91,31 @@ public static function get_status_list() {
...
+	 *
+	 * @param integer $course_id the course id.
+	 *
+	 * @return bool
+	 */
+	public static function is_course_accessible( int $course_id = 0 ) {
+		if ( ! $course_id || ! get_post( $course_id ) ) {
+			return false;
+		}
+
```
harunollyo, line 108: Here get the post in a variable and access the prop, later it reduce unnecessary db query.
```php
$course = get_post( $course_id );
if ( ! $course || ! is_object ( $course ) ) {
  return false;
}
```

models/CourseModel.php
```diff
@@ -91,6 +91,31 @@ public static function get_status_list() {
...
+		if ( ! in_array( get_post_type( $course_id ), array( tutor()->course_post_type, tutor()->free_course_post_type ), true ) ) {
```
harunollyo: Here check `$course->post_type`

ecommerce/CartController.php
```diff
@@ -219,12 +220,8 @@ public function add_course_to_cart() {
 		$user_id   = tutils()->get_user_id();
 		$course_id = Input::post( 'course_id', 0, Input::TYPE_INT );
```
harunollyo, line 221: Delete orphan cart items if course is delete by this hook
```php
add_action( 'deleted_post', array( $this, 'delete_cart_item' ) );
```

harunollyo (changes requested):

harunollyo (approved):

PR #2804, opened 2026-06-15
Use core api and utils from window instead of import

sazedul-haque:


assets/core/ts/index.ts
```diff
@@ -113,6 +124,43 @@ const initializePlugin = async () => {
...
+      TUTOR_CUSTOM_EVENTS,
+    },
+    config: {
+      tutorConfig: tutorConfig,
+    },
+    endpoints: endpoints,
+    query: queryServiceMeta.instance,
+    modal: modalServiceMeta.instance,
+    form: formServiceMeta.instance,
+    wpMedia: wpMediaServiceMeta.instance,
```
b-l-i-n-d: Those are already registered in the window.

b-l-i-n-d (changes requested):

assets/core/ts/index.ts
sazedul-haque: Removed

b-l-i-n-d (approved):

PR #2813, opened 2026-06-19
Improved security, corrected typos, and refactored code.

harunollyo:
Removed unused method
- tutor_default_settings


classes/Addons.php
```diff
@@ -179,8 +181,8 @@ public function addon_enable_disable() {
 
 		tutor_utils()->checking_nonce();
 
-		if ( ! current_user_can( 'manage_options' ) ) {
-			wp_send_json_error( array( 'message' => __( 'Access Denied', 'tutor' ) ) );
+		if ( ! User::is_admin() ) {
```
shewa12: We should check with the capability instead of the role. If a custom role has the manage_options capability then this check will fail.

classes/Course.php
```diff
@@ -3146,44 +3171,44 @@ public function course_enrollment() {
...
-				wp_send_json_error( __( 'Enrollment failed, please try again!', 'tutor' ) );
+		/**
+		 * This check was added to address a security issue where users could
+		 * enroll in a course via an AJAX call without purchasing it.
+		 *
+		 * To prevent this, we now verify whether the course is paid.
+		 * Additionally, we check if the user is already enrolled, since
+		 * Tutor's default behavior enrolls users automatically upon purchase.
+		 *
+		 * @since 3.9.4
```
shewa12, line 3163: The @since tag looks incorrect

classes/Lesson.php
```diff
@@ -536,12 +523,8 @@ public function ajax_delete_lesson() {
 
 		$lesson_id = Input::post( 'lesson_id', 0, Input::TYPE_INT );
 
-		if ( ! tutor_utils()->can_user_manage( 'lesson', $lesson_id ) ) {
-			$this->json_response(
-				tutor_utils()->error_message(),
-				null,
-				HttpHelper::STATUS_FORBIDDEN
-			);
+		if ( ! $lesson_id || ! tutor_utils()->can_user_manage( 'lesson', $lesson_id ) ) {
```
shewa12: Invalid lesson ID error & user permission error should have different responses.

classes/Lesson.php
```diff
@@ -617,6 +599,12 @@ public function mark_lesson_complete() {
 			return;
 		}
 
+		$course_id = tutor_utils()->get_course_id_by( 'lesson', $lesson_id );
+
+		if ( ! $course_id || ! tutor_utils()->is_enrolled( $course_id ) ) {
```
shewa12: Same here

shewa12 (changes requested):
Please check the comments

classes/Course.php
harunollyo: @since tag is correct. Code position changed

classes/User.php
```diff
@@ -145,6 +145,26 @@ public static function is( string $role ) {
...
+	 * User::can( 'edit_post_meta', $post->ID, $meta_key );
+	 *
+	 * @since 4.0.0
+	 *
+	 * @param string $capability capability.
+	 * @param mixed  ...$args     args.
+	 *
+	 * @return boolean
+	 */
+	public static function can( string $capability = 'manage_options', ...$args ) {
```
sazedul-haque, line 164: We should make the first argument mandatory to improve readability. A call like `User::can()` doesn't indicate which capability is being checked, making the code harder to understand at a glance.

harunollyo: What about tutor utils
```php
public function check_current_user_capability( string $capability = 'manage_options', int $object_id = 0 ) {

}

check_current_user_capability();
```

PR #2844, opened 2026-06-29
Fix RTL issues across dashboard and learning area

sazedul-haque:


b-l-i-n-d (approved):

components/SvgIcon.php
```diff
@@ -178,6 +185,23 @@ public function ignore_kids( bool $ignore_kids = true ): self {
...
+	 * Use this only for directional icons such as arrows,
+	 * chevrons and pagination controls.
+	 *
+	 * @since 4.0.0
+	 *
+	 * @param bool $flip Whether to flip in RTL.
+	 *
+	 * @return $this
+	 */
+	public function flip_rtl( bool $flip = true ): self {
```
harunollyo: I think the method should be named something like:

```php
->rtl_aware()->render();
->rtl()->render();
```

and should not require any parameters. The parameter doesn't affect the behavior since everything is handled through CSS based on the `dir="rtl"` attribute on the HTML element. The method only needs to mark the component as RTL-aware, while the actual flipping/adaptation is handled entirely by CSS.

components/Button.php
```diff
@@ -257,6 +266,23 @@ public function icon_only( bool $icon_only = true ): self {
...
+	 * Use this only for directional icons such as arrows,
+	 * chevrons and pagination controls.
+	 *
+	 * @since 4.0.0
+	 *
+	 * @param bool $flip Whether to flip the icon in RTL.
+	 *
+	 * @return $this
+	 */
+	public function flip_icon_rtl( bool $flip = true ): self {
```
harunollyo: I think the method should be named something like:

```php
->rtl_aware()->render();
->rtl()->render();
```

and should not require any parameters. The parameter doesn't affect the behavior since everything is handled through CSS based on the `dir="rtl"` attribute on the HTML element. The method only needs to mark the component as RTL-aware, while the actual flipping/adaptation is handled entirely by CSS.

harunollyo (changes requested):

components/SvgIcon.php
sazedul-haque: I have removed the parameter. Every component in our system is RTL aware so naming the method `rtl_aware` does not specify what will happen. It's too vague. So, I am keeping the `flip_rtl` name.

components/Button.php
sazedul-haque: Renamed it to `flip_rtl` to make unified.

PR #2851, opened 2026-07-01
Hide course and lesson complete buttons if not enrolled

sazedul-haque:


views/modal/review.php
```diff
@@ -13,6 +13,8 @@
 use Tutor\Options_V2;
 
 $is_legacy_learning = Options_V2::LEARNING_MODE_LEGACY === tutor_utils()->get_option( 'learning_mode' );
+
+$modal_id = 'tutor-review-modal-' . uniqid();
```
harunollyo: Use `$course_id` here instead of `uniqid()`

harunollyo (changes requested):

views/modal/review.php
sazedul-haque: Updated

harunollyo (approved):

PR #2897, opened 2026-07-10
Automatic result calculation for Image answering type quiz

harunollyo:
Currently, the Image Answering quiz type is treated as a manual review question. However, when creating this quiz type, instructors are required to define the correct answer, which means the system already has the necessary information to evaluate submissions automatically.

This change updates the behavior so that Image Answering quizzes are graded automatically based on the stored correct answer, eliminating the need for manual review for this quiz type.

### DIFF
--- a/classes/Quiz.php
+++ b/classes/Quiz.php
@@ -899,42 +899,32 @@ class Quiz {
 							$is_answer_was_correct = true;
 						}
 					} elseif ( QuizModel::QUESTION_TYPE_IMAGE_ANSWERING === $question_type ) {
-						$image_inputs          = tutor_utils()->avalue_dot( 'answer_id', $answers );
-						$image_inputs          = (array) array_map( 'sanitize_text_field', $image_inputs );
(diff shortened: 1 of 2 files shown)


classes/Quiz.php
```diff
@@ -899,42 +899,32 @@ function ( $ans ) {
...
+								QuizModel::QUESTION_TYPE_IMAGE_ANSWERING
+							)
+						);
+
+						$is_answer_was_correct = true;
+						foreach ( $stored_answers as $stored_answer ) {
+							$user_answer    = isset( $image_inputs[ $stored_answer->answer_id ] ) ? trim( wp_unslash( $image_inputs[ $stored_answer->answer_id ] ) ) : '';
+							$correct_answer = trim( wp_unslash( $stored_answer->answer_title ) );
+
+							if ( 0 !== strcasecmp( $user_answer, $correct_answer ) ) {
```
shewa12, line 923: Why compare in a case-insensitive manner?

shewa12 (approved):

PR #2901, opened 2026-07-14
🐛 Fix disabled nav items interaction, scope body reset, add pnpm workspace

b-l-i-n-d:
- Removes aria-disabled="true" from disabled nav items in quiz and lesson navigation
- Removes conditional href="#" fallback for disabled nav items
- Removes pointer-events: none CSS rule on disabled sidebar items
- Adds pnpm-workspace.yaml configuration for package manager builds
- Scopes body CSS reset under [data-tutor-theme] attribute selector

https://app.clickup.com/t/9018365849/86ey98yx1

### DIFF
--- a/classes/Quiz.php
+++ b/classes/Quiz.php
@@ -2268,10 +2268,9 @@ class Quiz {
 		?>
 
 		<a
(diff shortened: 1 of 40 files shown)


classes/Quiz.php
shewa12: Why the # removed?

shewa12 (approved):

PR #2907, opened 2026-07-16
🐛 [QUIZ] Use array index for ordering display and correct dnd review data-option

b-l-i-n-d:
- Fixed ordering question display to use array index instead of stored answer_order value
- Fixed drag-and-drop review data-option attribute to use given_status instead of given_text


templates/shared/components/quiz/attempt-details/questions/review-answer-dnd.php
```diff
@@ -115,7 +115,7 @@
 			$correct_text = isset( $row['correct_text'] ) ? wp_unslash( $row['correct_text'] ) : '';
 			?>
 			<div class="tutor-quiz-review-dnd-row">
-				<div class="tutor-quiz-review-item tutor-quiz-review-given" data-option="<?php echo esc_attr( $given_text ); ?>">
+				<div class="tutor-quiz-review-item tutor-quiz-review-given" data-option="<?php echo esc_attr( $row['given_status'] ); ?>">
```
shewa12: Check if the $row is an array and has a given_status element before accessing it
`<?php echo esc_attr( $row['given_status'] ); ?>`

shewa12 (changes requested):

PR #2929, opened 2026-07-27
🛠️ Dashbord settings UI issue on mobile device

b-l-i-n-d:
- Resolve resize event trigger 
- Resolve nested overflow container

https://app.clickup.com/t/9018365849/86eyddu21

### DIFF
--- a/templates/dashboard/account/settings.php
+++ b/templates/dashboard/account/settings.php
@@ -79,6 +79,7 @@ $settings_tab_data = array_values(
 <section x-data="tutorSettings()">
 	<div 
 		x-data='(() => {
+			const initialTab = new URL(window.location.href).searchParams.get("tab");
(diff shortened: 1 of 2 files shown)


templates/dashboard/account/settings.php
```diff
@@ -123,20 +142,7 @@ class="tutor-profile-settings-section"
...
-					}
-				})"
-				@resize.window="
-					if (window.innerWidth >= 768 && activeTab === 'none') {
-						selectTab('account');
-					} else if (window.innerWidth < 768 && activeTab !== 'none') {
-						selectTab('none');
-					}
-				"
+				@resize.window="handleViewportResize()"
```
sazedul-haque, line 145: We should remove this window resize option all together. It has no real use case.

shewa12 (approved):

PR #2936, opened 2026-07-30
 ✨ Add iframe suppprt to course description

b-l-i-n-d:
- Add iframe support to course description
- Resolve all phpstan warnings

https://app.clickup.com/t/9018365849/86eyda2jf

### DIFF
--- a/classes/Course.php
+++ b/classes/Course.php
@@ -208,7 +208,12 @@ class Course extends Tutor_Base {
 		 *
 		 * @since v.1.6.6
 		 */
-		add_action( 'deleted_post', array( new CourseModel(), 'delete_course_data' ) );
+		add_action(
+			'deleted_post',
+			function ( $post_id ) {
+				( new CourseModel() )->delete_course_data( $post_id );
+			}
+		);
 
 		/**
 		 * Delete course data after deleted course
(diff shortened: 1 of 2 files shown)


classes/Course.php
```diff
@@ -208,7 +208,12 @@ public function __construct( $register_hooks = true ) {
 		 *
 		 * @since v.1.6.6
 		 */
-		add_action( 'deleted_post', array( new CourseModel(), 'delete_course_data' ) );
+		add_action(
```
shewa12, line 211: We don't need to wrap the callback inside a fn to receive the param. WordPress resolves it automatically.

classes/Course.php
```diff
@@ -355,12 +360,12 @@ public function handle_password_submit() {
...
 	 *
-	 * @param bool    $required Whether the password is required.
-	 * @param WP_Post $post     The post object.
+	 * @param bool     $required Whether the password is required.
+	 * @param \WP_Post $post     The post object.
 	 *
 	 * @return bool false if the current user is enrolled, original value otherwise.
 	 */
-	public function bypass_password_for_enrolled( $required, $post ) {
+	public function bypass_password_for_enrolled( $required, \WP_Post $post ) {
```
shewa12: Why was type-hint added? If we want to add a type hint, then we need to make sure the calling scripts must pass a WP_Post instance.

shewa12 (approved):

PR #2947, opened 2026-08-05
🔒 Filter assigned course ids by instructor status

b-l-i-n-d:
- Add an optional `$status` parameter to `get_assigned_courses_ids_by_instructors()` to filter assigned course IDs by instructor status.
- Restrict the submitted-assignment query to courses where the user is an approved co-instructor.
- Keep existing callers unaffected when no status is provided.

### DIFF
--- a/classes/Utils.php
+++ b/classes/Utils.php
@@ -6317,37 +6317,29 @@ class Utils {
 		global $wpdb;
 		$user_id = $this->get_user_id( $user_id );
 
+		$where_clause = "meta.meta_key = '_tutor_instructor_course_id' AND meta.user_id = %d";
+		$args         = array( $user_id );
+
 		if ( ! empty( $status ) ) {
-			$query = $wpdb->prepare(
-				"SELECT meta.meta_value
-				FROM {$wpdb->usermeta} meta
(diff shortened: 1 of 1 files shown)


classes/Utils.php
```diff
@@ -6306,27 +6306,50 @@ public function get_assignments() {
...
 	 * @return array
 	 */
-	public function get_assigned_courses_ids_by_instructors( $user_id = 0 ) {
+	public function get_assigned_courses_ids_by_instructors( $user_id = 0, $status = '' ) {
 		global $wpdb;
 		$user_id = $this->get_user_id( $user_id );
 
-		$get_assigned_courses_ids = $wpdb->get_col(
-			$wpdb->prepare(
+		if ( ! empty( $status ) ) {
```
shewa12, line 6323: Instead of keeping the same query in the if & else block, just make the where condition dynamic and pass in the sql query.

This would be come
```
WHERE meta_status.user_id = meta.user_id
							AND meta_status.meta_key = '_tutor_instructor_status'
							AND meta_status.meta_value = %s
```
This 
```
Where {$where_clause}
```
Check existing codebases; many references are there!

shewa12 (changes requested):

shewa12 (approved):

PR #2959, opened 2026-08-14
Security Issues

shewa12:
3 different security issues have been fixed. [See Details](https://ollyo.slack.com/archives/C02A30318PQ/p1786535695920049)

### DIFF
--- a/classes/Utils.php
+++ b/classes/Utils.php
@@ -1547,6 +1547,38 @@ class Utils {
 	}
 
 	/**
+	 * Whether a scalar looks like an absolute local filesystem path.
+	 *
+	 * Used when filtering video meta so paths cannot be smuggled under another key.
+	 * Does not call file_exists() — that would allow path probing on save.
+	 *
+	 * @since 4.0.6
+	 *
+	 * @param mixed $value value to inspect.
+	 *
+	 * @return bool
+	 */
+	private function is_local_filesystem_path_value( $value ) {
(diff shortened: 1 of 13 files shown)


classes/Utils.php
```diff
@@ -1542,10 +1542,59 @@ public function update_video( $post_id = 0, $video_data = array() ) {
...
+				'poster_url',
+				'runtime',
+				'duration_sec',
+				'playtime',
+			)
+		);
+
+		$filtered = array();
+		foreach ( $video as $key => $value ) {
+			if ( 'path' === $key ) {
```
harunollyo, line 1625: If user pass `$value` as path, can we rely on key name?
```php
if ( ( str_contains($value, '/') || str_contains($value, '\\') ) && file_exists( $value ) ) {
 continue;
}
```

harunollyo (changes requested):

classes/Utils.php
shewa12: Good catch — we shouldn't rely on the key name only. Checking 
`str_contains('/') && file_exists()` would strip valid YouTube/CDN URLs and enable path probing on save. 

We keep the allow-list, still drop path, and additionally reject scalar values that look like absolute local filesystem paths **(while allowing http(s)://).** 
Streaming path still comes only from the attachment in `get_video_info()`, with an uploads-dir check in Video_Stream.

PR #2978, opened 2026-08-26
💅Update color tokens

b-l-i-n-d:
https://app.clickup.com/t/9018365849/z8zyy1chcg

### DIFF
--- a/classes/Assets.php
+++ b/classes/Assets.php
@@ -519,8 +519,8 @@ class Assets {
 		$admin_colors = array();
 		if ( is_admin() ) {
 			$admin_colors = array(
-				'--tutor-color-primary'       => '#3E64DE',
-				'--tutor-color-primary-hover' => '#3A5CCC',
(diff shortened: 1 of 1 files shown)


classes/Assets.php
```diff
@@ -519,17 +519,17 @@ private function load_color_palette() {
 		$admin_colors = array();
 		if ( is_admin() ) {
 			$admin_colors = array(
-				'--tutor-color-primary'       => '#3E64DE',
-				'--tutor-color-primary-hover' => '#3A5CCC',
+				'--tutor-color-primary'       => '#3e64de',
```
sazedul-haque: Primary color looks same as previous one. why the change?

sazedul-haque (approved):

classes/Assets.php
b-l-i-n-d: Just short for of that color code, maybe caused while copying from color tokens. Made all of them capital.

PR #2980, opened 2026-08-26
Fix: `Import From Course` in content bank not showing correct count for instructors

Anindra123:
https://app.clickup.com/t/9018365849/86eyjzj50


classes/Instructor.php
```diff
@@ -100,6 +100,36 @@ public function __construct( $register_hook = true ) {
 
 		add_action( 'wp_ajax_tutor_save_instructor_home_sections_order', array( $this, 'ajax_save_home_sections_order' ) );
 		add_action( 'wp_ajax_tutor_save_instructor_home_sections_visibility', array( $this, 'ajax_save_home_section_visibility' ) );
+		add_filter( 'map_meta_cap', array( $this, 'map_meta_cap_for_instructor' ), 10, 4 );
```
shewa12: We don't need this filter; instead, add capability if the current user is an instructor and doesn't have the capability.

migrations/InstructorCapabilityMigrator.php
```diff
@@ -0,0 +1,132 @@
...
+	 * @param int $offset offset.
+	 * @param int $limit limit.
+	 *
+	 * @return array
+	 */
+	protected function get_items( $offset, $limit ): array {
+		$this->user_model = new UserModel();
+		$users            = $this->user_model->get_users_list(
+			array(
+				'role'   => tutor()->instructor_role,
```
shewa12: We can also add status check to fetch only the approved instructors.

migrations/InstructorCapabilityMigrator.php
```diff
@@ -0,0 +1,132 @@
...
+	 *
+	 * @since 4.0.8
+	 *
+	 * @return int
+	 */
+	protected function get_total_items(): int {
+		$this->user_model = new UserModel();
+		$users            = $this->user_model->get_users_list(
+			array(
+				'role'   => tutor()->instructor_role,
```
shewa12: We can also add status check to fetch only the approved instructors.

shewa12 (approved):

PR #2982, opened 2026-08-27
Refactored the course completion rate db query & implemented cache to load page faster

shewa12:
Previously, this method: `get_course_completion_distribution_data_by_instructor ` was executing an `N + 1` query. Inside the loop, calling other helper methods was executing more queries that was blocking the page load when the number of enrollments are high.

Refactored the entire query and implemented caching.

[Related PRO PR](https://github.com/themeum/tutor-pro/pull/1900)

### DIFF
--- a/classes/Instructor.php
+++ b/classes/Instructor.php
@@ -17,6 +17,7 @@ use DateTime;
 use Tutor\GDPR\Controllers\LegalConsent;
 use Tutor\Helpers\DateTimeHelper;
 use Tutor\Helpers\QueryHelper;
+use Tutor\Models\CourseModel;
 use Tutor\Traits\JsonResponse;
 
 /**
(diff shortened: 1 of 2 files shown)


classes/Instructor.php
```diff
@@ -507,60 +536,233 @@ public static function get_course_completion_distribution_data_by_instructor( $i
...
+			$topic_post_parent_clause      = "AND topic.post_parent IN ({$placeholders})";
+			$enrollment_post_parent_clause = "AND e.post_parent IN ({$placeholders})";
+		}
 
-			// Completed course.
-			if ( tutor_utils()->is_completed_course( $enrollment->post_parent, $enrollment->post_author ) ) {
-				++$counts['completed'];
-				continue;
-			}
+		$sql = "
```
sanjana4khan, line 588: Need to add a condition for instructors with zero published courses, as they are currently able to see overall admin data.

sanjana4khan (changes requested):

classes/Instructor.php
shewa12: Added

sanjana4khan (approved):

PR #2985, opened 2026-09-01
🚀 Instructor dashboard concurrent ajax lazyloading & zero-cls skeletons

b-l-i-n-d:
- Add granular concurrent AJAX lazy-loading for instructor dashboard sections
- Add zero-CLS Skeleton wireframes with matched geometry across desktop and mobile
- Introduce `InstructorMetricsAdapter` to decouple domain metrics calculations from presentation
- Introduce `DashboardSectionManager` to handle dashboard section dispatching, hooks, and AJAX endpoints
- Add `tutorLazySection` Alpine component with query caching, error toasts, and visibility observers
- Centralize custom event constants (`TUTOR_CUSTOM_EVENTS`) across date filter and sorting listeners
- Resolve query parameter clobbering by namespacing top performing courses filter to `top_performing_course`
- Eliminate DateFilter clear button initial load layout flash
- Add architectural decision records (`0001-dashboard-ajax-lazyloading-architecture.md`, `0002-dashboard-metrics-adapter-pattern.md`)


classes/DashboardSectionManager.php
```diff
@@ -0,0 +1,528 @@
...
+		$params = array(
+			'user_id'               => $user_id,
+			'start_date'            => $start_date,
+			'end_date'              => $end_date,
+			'top_performing_course' => $type,
+			'type'                  => $type,
+		);
+
+		try {
+			$result = self::render_section( $section_id, $params );
```
shewa12: This method is not throwing an exception. Try-catch is not useful here.

classes/DashboardSectionManager.php
```diff
@@ -0,0 +1,528 @@
...
+	 * Render Current Stats View
+	 *
+	 * @param array $data   Stats cards from adapter.
+	 * @param array $params Context parameters.
+	 * @return array
+	 *
+	 * @since 4.0.9
+	 */
+	protected static function render_current_stats( $data, array $params ): array {
+		$start_date = $params['start_date'] ?? '';
```
shewa12: How are we sanitizing the data?

classes/DashboardSectionManager.php
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	/**
+	 * Render Overview Chart View
+	 *
+	 * @param array $data   Chart data from adapter.
+	 * @param array $params Context parameters.
+	 * @return array
+	 *
+	 * @since 4.0.9
```
shewa12: Check all the docs comments, the format is not what we use. @since tag should on top of param, one line gap should be added before any different annotation.

classes/InstructorMetricsAdapter.php
```diff
@@ -0,0 +1,357 @@
...
+	 * Date range array helper
+	 *
+	 * @param string $from Start date.
+	 * @param string $to   End date.
+	 * @return array
+	 *
+	 * @since 4.0.9
+	 */
+	public static function date_range( string $from, string $to ): array {
+		return array(
```
shewa12, line 40: Sanitization missing

assets/core/ts/components/calendar.ts
```diff
@@ -553,18 +592,40 @@ export function calendar({ options, hidePopover }: { options: Options; hidePopov
     applyPreset(preset: Preset) {
       if (!this.calendar) return;
 
+      hidePopover?.();
+
       const dates = this.getPresetDates(preset);
+      const presetLabels: Record<string, string> = {
```
sazedul-haque: These strings is already defined  at top. why duplicate?

classes/DashboardSectionManager.php
```diff
@@ -0,0 +1,532 @@
...
+	 * Handle AJAX request to lazyload a dashboard section.
+	 *
+	 * @since 4.0.8
+	 *
+	 * @return void
+	 */
+	public function ajax_get_dashboard_section() {
+		tutor_utils()->check_nonce();
+
+		if ( ! User::is_admin() && ! tutor_utils()->is_instructor() ) {
```
sazedul-haque, line 211: Admin might not be an instructor

assets/core/ts/components/calendar.ts
b-l-i-n-d: It is an unintentional redundancy 🥲

classes/DashboardSectionManager.php
b-l-i-n-d: In that case, the user will not enter this block.

sazedul-haque: Yes but admin should be able to access the dashboard data. Right?

b-l-i-n-d: Access is granted if the user is either an admin or an instructor.

sazedul-haque: Got it.

shewa12 (approved):

PR #3006, opened 2026-09-15
✨ Partial and negative marking support for quizzes

b-l-i-n-d:
- Consolidates question feedback and manual override map extraction into shared `QuizModel` helpers
- Ends quiz attempts that reach a graded state after manual review
- Marks attempts as pending correctly when a manual-review answer has not been graded yet
- Standardizes correct/incorrect answer status flags via `QuizModel` constants
- Extracts bulk manual mark and question feedback application during manual quiz review
- Enables admin toggles to complete turn-off flow when the confirm-usage check passes
- Cleans up template feedback lookup to use shared helpers


models/QuizModel.php
```diff
@@ -1094,18 +1108,139 @@ public static function get_attempt_answer_status( $attempt_answer ): string {
...
+	 * Fully correct answers are counted under 'correct' and incorrect answers under 'incorrect'.
+	 * Pending, graded, and skipped answers are excluded.
+	 *
+	 * @since 4.1.0
+	 *
+	 * @param array|null $answers List of answer objects.
+	 *
+	 * @return array Associative array with answer status counts.
+	 */
+	public static function get_attempt_answer_counts( $answers ): array {
```
shewa12, line 1227: This method only checks whether it is correct or incorrect; what about graded? If the question types are open-ended or short-answer, then it is graded, not marked correct or incorrect.

classes/Quiz.php
```diff
@@ -1251,14 +1406,172 @@ private function review_quiz_answers_bulk( int $attempt_id, array $review_status
...
+		$answer_updated = QueryHelper::update(
+			'tutor_quiz_attempt_answers',
+			array(
+				'achieved_mark' => $new_mark,
+				'is_correct'    => QuizModel::ATTEMPT_ANSWER_MANUAL_GRADED,
+			),
+			array( 'attempt_answer_id' => (int) $attempt_answer->attempt_answer_id )
+		);
+
+		if ( ! $answer_updated ) {
```
shewa12: Use false === $answer_updated. If the user saves the same as the previous data then WP Update will return 0, and it will return null, which will generate the wrong mark.

assets/src/js/v3/entries/course-builder/components/modals/QuizModal.tsx
```diff
@@ -101,6 +102,10 @@ const QuizModal = ({
         hide_question_number_overview: false,
         short_answer_characters_limit: 200,
         open_ended_answer_characters_limit: 500,
+        enable_partial_marking: false,
+        enable_negative_marking: false,
+        negative_mark_type: tutorConfig.settings?.quiz_negative_mark_type === 'fixed' ? 'fixed' : 'percent',
```
sazedul-haque: Use constant for the types.

assets/src/js/v3/entries/course-builder/components/modals/QuizModal.tsx
```diff
@@ -101,6 +102,10 @@ const QuizModal = ({
...
+        negative_mark_value: Number(tutorConfig.settings?.quiz_negative_mark_value ?? 0.15),
```
sazedul-haque: Where did `0.15` come from? use constant.

assets/src/js/v3/entries/course-builder/services/quiz.ts
```diff
@@ -102,6 +102,10 @@ export interface QuizDetailsResponse {
     hide_question_number_overview: '0' | '1';
     short_answer_characters_limit: number;
     open_ended_answer_characters_limit: number;
+    enable_partial_marking?: '0' | '1';
+    enable_negative_marking?: '0' | '1';
+    negative_mark_type?: 'percent' | 'fixed';
```
sazedul-haque: Use constant here and other places in this file.

assets/src/js/v2/quiz-attempt.js
```diff
@@ -65,4 +70,172 @@ window.addEventListener('DOMContentLoaded', function() {
...
+    // Save or delete feedback via AJAX with shared loading state and error toast.
+    const persistFeedback = async (action, button) => {
+        const { attemptInput, answerInput, textarea } = feedbackModalElements();
+        const attemptId       = attemptInput.value;
+        const attemptAnswerId = answerInput.value;
+        const feedback        = textarea.value.trim();
+
+        if (!attemptId || !attemptAnswerId) return null;
+
+        const formData = new FormData();
```
sazedul-haque: User `tutorFormData` from helpers.

classes/Quiz.php
```diff
@@ -1192,30 +1343,34 @@ public function review_quiz_answer() {
 	/**
 	 * Review quiz answers in bulk for v4 dashboard flow.
 	 *
-	 * @since 4.0.0
+	 * @since 4.1.0
```
sazedul-haque: Why version change when this was introduced in 4.0.0?

classes/Quiz.php
```diff
@@ -1192,30 +1343,34 @@ public function review_quiz_answer() {
...
 
-		$this->review_quiz_answers_bulk( $attempt_id, $review_statuses );
+		$this->review_quiz_answers_bulk( $attempt_id, $review_statuses, $manual_marks, $question_feedback );
 	}
 
 	/**
 	 * Review quiz answers in bulk for v4 dashboard flow.
 	 *
-	 * @since 4.0.0
+	 * @since 4.1.0
```
sazedul-haque: Same

classes/Quiz_Attempts_List.php
```diff
@@ -641,6 +641,70 @@ public static function render_quiz_attempt_marks_percentage( $attempt_result = '
...
+	 * @return array
+	 */
+	public static function get_quiz_attempt_summary_statics( $attempt_data, $answers ) {
+		$answer_counts = QuizModel::get_attempt_answer_counts( $answers );
+
+		$static_items = array(
+			'correct'   => array(
+				'class' => 'correct',
+				/* translators: %d: number of correct answers. */
+				'label' => __( '<span class="tutor-font-semibold tutor-text-primary">%d</span> correct', 'tutor' ),
```
sazedul-haque: Why translating with the markup? All of them seems to have same classes.

classes/Quiz.php
```diff
@@ -1251,14 +1406,172 @@ private function review_quiz_answers_bulk( int $attempt_id, array $review_status
...
+			$question_mark = (float) ( $question->question_mark ?? 0.0 );
+		}
+
+		$new_mark       = min( max( 0, (float) $mark ), $question_mark );
+		$previous_mark  = (float) ( $attempt_answer->achieved_mark ?? 0.0 );
+		$answer_updated = QueryHelper::update(
+			'tutor_quiz_attempt_answers',
+			array(
+				'achieved_mark' => $new_mark,
+				'is_correct'    => QuizModel::ATTEMPT_ANSWER_MANUAL_GRADED,
```
sazedul-haque, line 1481: Shouldn't we maintain a different flag like `is_manually_graded` for this purpose?

templates/shared/components/quiz/attempt-details/questions/open-ended.php
```diff
@@ -23,17 +31,193 @@
...
+					?>
+
+					<div
+						class="tutor-quiz-feedback-panel"
+						x-show="expanded"
+						x-collapse
+						x-cloak
+					>
+						<div class="tutor-quiz-feedback-panel-header">
+							<span class="tutor-quiz-feedback-panel-title" x-text="feedback ? <?php echo esc_attr( wp_json_encode( __( 'Edit Feedback', 'tutor' ) ) ); ?> : <?php echo esc_attr( wp_json_encode( __( 'Write feedback', 'tutor' ) ) ); ?>"></span>
```
sazedul-haque: wp_json_encode may not needed here.

sazedul-haque (changes requested):

classes/Quiz.php
b-l-i-n-d: This was an architectural decision.

models/QuizModel.php
b-l-i-n-d: We currently only require the counts for `correct` and `incorrect`; no other counts are needed.

PR #3007, opened 2026-09-16
Fixed Various Security Issues

shewa12:
Check the links below to understand the security issues:
[Issue 1](https://ollyo.slack.com/archives/C02A30318PQ/p1789099910883099)
[Issue 2](https://ollyo.slack.com/archives/C02A30318PQ/p1789099941523009)
[Issue 3](https://ollyo.slack.com/archives/C02A30318PQ/p1789099970693019)
[Issue 4](https://ollyo.slack.com/archives/C02A30318PQ/p1789278941956839)


sazedul-haque (approved):

templates/dashboard/account/settings.php
```diff
@@ -92,7 +92,7 @@ function ( $tab ) {
 
 			return {
 				...tabs,
-				backUrl: <?php echo wp_json_encode( $back_url ); ?>,
+				backUrl: <?php echo wp_json_encode( $back_url, JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_TAG | JSON_HEX_AMP ); ?>,
```
harunollyo, line 95: We don't need `wp_json_encode` for URL. simply we can use `esc_url( $back_url )`

templates/dashboard/account/settings.php
```diff
@@ -81,7 +81,7 @@ function ( $tab ) {
 		x-data='(() => {
 			const initialTab = new URL(window.location.href).searchParams.get("tab");
 			const tabs = tutorTabs({
-				tabs: <?php echo wp_json_encode( $settings_tab_data ); ?>,
+				tabs: <?php echo wp_json_encode( $settings_tab_data, JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_TAG | JSON_HEX_AMP ); ?>,
```
harunollyo, line 84: Instead of passing flag, can we use escaping function `esc_attr` or `esc_js`?
```php
tabs: <?php echo esc_js( wp_json_encode( $settings_tab_data ) ); ?>
```

templates/user-profile.php
```diff
@@ -1,4 +1,6 @@
 <?php
+
+use TUTOR\User;
```
harunollyo: I think, this import statement unintentionally misplaced.

views/pages/ecommerce/order-list.php
```diff
@@ -1,4 +1,6 @@
 <?php
+
+use TUTOR\User;
```
harunollyo: same here

harunollyo (changes requested):

templates/dashboard/account/settings.php
shewa12: Why don't we need wp_json_encode? Here these flags: `JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_TAG | JSON_HEX_AMP`
Explicitly Unicode-encoding the characters that browsers parse as an attribute. The `wp_json_encode` makes sure the string type and flag secure the rest.

templates/dashboard/account/settings.php
shewa12: In this case, esc_js or attr is not a proper fix.

